PR #1, opened 2024-09-05
Databricks client implementation

serena-ruan:
Initial version of UC function client
- BaseFunctionClient as the basic client
- DatabricksFunctionClient for connecting with databricks managed UC

Main functionalities:
- create_function
- get_function
- list_functions
- execute_function

Tests added in test_client function can be executed; test_databricks_client is tested locally due to the limitation of authentication and warehouse usage; 

NOTE: create_function with function_info is pending on a bug on the API https://docs.databricks.com/api/workspace/functions/create, so please ignore this part for now.


tests/functions/test_databricks_client.py
```diff
@@ -0,0 +1,720 @@
...
+"""
+    return FunctionInputOutput(
+        sql_body=sql_body,
+        func_name=f"{CATALOG}.{SCHEMA}.{func_name}",
+        inputs=[{"start": datetime.date(2024, 1, 1), "end": "2024-01-07"}],
+        output="day_of_week,day\n1,2024-01-01\n2,2024-01-02\n3,2024-01-03\n4,2024-01-04\n5,2024-01-05\n",
+    )
+
+
+def test_create_function_with_function_info(client: DatabricksFunctionClient):
```
serena-ruan: Pls ignore this test for now as it's blocked by an API bug

pyproject.toml
```diff
@@ -0,0 +1,60 @@
+[project]
+name = "unitycatalog-function"
+version = "0.0.1"
+description = "Official python library for unitycatalog functions support"
+authors = [
+    { name="Serena Ruan", email="serena.ruan@databricks.com" }
+]
+readme = "README.md"
+requires-python = ">=3.8"
```
harupy:
```suggestion
requires-python = ">=3.9"
```

3.8 will reach EOL soon.

pyproject.toml
```diff
@@ -0,0 +1,60 @@
+[project]
+name = "unitycatalog-function"
```
harupy:
```suggestion
name = "unitycatalog-functions"
```

should this be plural?

harupy: btw is this a name space package?

serena-ruan: Yea this name is pending to be changed, I'm waiting for Xiangrui's response :)

README.md
```diff
@@ -1 +1,72 @@
 ## Unitycatalog function library
+
+The Unitycatalog function library provides convenient APIs to interact with Unitycatalog functions, including functions creation, retrieval and execution.
+The library includes clients for interacting with both Unitycatalog OSS server and Databricks-managed Unitycatalog.
+
+## Installation
+
+```sh
+# install from the production repo
```
harupy: what is a production repo?

pyproject.toml
serena-ruan: lol just found it's Oct 2024!

README.md
serena-ruan: Maybe just 'repo' 😆

unitycatalog/functions/client.py
harupy, line 1: This is not a blocker, but I'd recommend using the `src` layout. https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/ illustrates why the src layout makes sense.

pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+version = "0.0.1"
+description = "Official python library for unitycatalog functions support"
```
harupy:
```suggestion
description = "Official python library for Unity Catalog functions support"
```

pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+
+
+[project.optional-dependencies]
+dev = [
+  "pytest",
+  "databricks-sdk",
+  # Note: databricks-connect 15.1 is recommended for serverless, databricks-connect>=14.0.0 requires python>=3.10
+  "databricks-connect==15.1.0",
+  "pandas",
+  "ruff",
```
harupy:
```suggestion
  "ruff==x.y",
```

let's pin ruff.

README.md
harupy: I'd say "Install from the source"

unitycatalog/functions/client.py
```diff
@@ -0,0 +1,136 @@
...
+    def __init__(self, client: Any = None, **kwargs: Any) -> None:  # noqa: ARG002
+        self._lock = threading.Lock()
+
+    @abstractmethod
+    def create_function(self, function_info: Any) -> Any:
+        """Create a function"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def retrieve_function(self, function_name: str, **kwargs: Any) -> Any:
```
harupy:
```suggestion
    def get_function(self, function_name: str, **kwargs: Any) -> Any:
```

for easier and faster typing.

README.md
```diff
@@ -1 +1,72 @@
...
+## Get started
+
+### Databricks-managed UC
+To use Databricks-managed Unitycatalog, please follow the [instruction](https://docs.databricks.com/en/dev-tools/cli/authentication.html#authentication-for-the-databricks-cli) to authenticate first and make sure the token has workspace-level privilege for managing UC functions.
+
+#### Prerequisites
+- Install databricks-sdk package with `pip install databricks-sdk`.
+- For accessing the UC functions in Databricks, please create a SQL warehouse following [this instruction](https://docs.databricks.com/en/compute/sql-warehouse/create.html), and save the warehouse id.
+- [Optional] Install databricks-connect package with `pip install databricks-connect`.
+    - This package is only required for creating UC functions using sql body.
```
harupy, line 25: Is `databricks-sdk` lacking this functionality?

unitycatalog/functions/client.py
```diff
@@ -0,0 +1,136 @@
...
+                of type InputParams in unitycatalog, or FunctionParameterInfos in databricks.
+            parameters (Dict[str, Any]): The parameters to validate.
+        """
+        parameters = deepcopy(parameters)
+        if input_params and input_params.parameters:
+            invalid_params: Dict[str, str] = {}
+            for param in input_params.parameters:
+                if param.name in parameters:
+                    try:
+                        self._validate_param_type(parameters.pop(param.name), param)
```
harupy: do we need to use pop? can we use get?

unitycatalog/functions/client.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @abstractmethod
+    def _execute_uc_function(
+        self, function_info: Any, parameters: Dict[str, Any], **kwargs: Any
+    ) -> FunctionExecutionResult:
+        raise NotImplementedError
+
+
+# TODO: update BaseFunctionClient to Union[BaseFunctionClient, AsyncBaseFunctionClient] after async client is supported
+def get_uc_function_client() -> Optional[BaseFunctionClient]:
```
harupy: how is this function used?

unitycatalog/functions/databricks.py
```diff
@@ -0,0 +1,549 @@
...
+    except ImportError as e:
+        raise ImportError(
+            "Could not import databricks-sdk python package. "
+            "If you want to use databricks backend then "
+            "please install it with `pip install databricks-sdk`."
+        ) from e
+    return WorkspaceClient()
+
+
+def column_type_to_python_type(column_type: Enum) -> Any:
```
harupy:
```suggestion
def column_type_to_python_type(column_type: ColumnTypeName) -> Any:
```

unitycatalog/functions/client.py
```diff
@@ -0,0 +1,136 @@
...
+    Base class for uc function calling client
+    """
+
+    def __init__(self, client: Any = None, **kwargs: Any) -> None:  # noqa: ARG002
+        self._lock = threading.Lock()
+
+    @abstractmethod
+    def create_function(self, function_info: Any) -> Any:
+        """Create a function"""
+        raise NotImplementedError
```
harupy:
```suggestion
```

can this be removed?

unitycatalog/functions/databricks.py
```diff
@@ -0,0 +1,549 @@
...
+        self, function_name: str, **kwargs: Any
+    ) -> Union["FunctionInfo", List["FunctionInfo"]]:
+        splits = function_name.split(".")
+        if len(splits) != 3:
+            raise ValueError(
+                f"Invalid function name: {function_name}, expecting format <catalog_name>.<schema_name>.<function_name>."
+            )
+        if splits[-1] == "*":
+            functions = self.client.functions.list(catalog_name=splits[0], schema_name=splits[1])
+            return list(functions)
```
harupy: Can we add a `list_functions` instead?

unitycatalog/functions/databricks.py
```diff
@@ -0,0 +1,549 @@
...
+            raise ValueError(
+                f"Parameter {param_info.name} should be of type {param_info.type_name.value} "
+                f"(corresponding python type {value_python_type}), but got {type(value)}"
+            )
+        validate_param(value, param_info)
+
+    @override
+    def execute_function(
+        self, function_name: str, parameters: Optional[Dict[str, Any]] = None, **kwargs: Any
+    ) -> FunctionExecutionResult:
```
harupy: do we need to override this?

unitycatalog/functions/databricks.py
```diff
@@ -0,0 +1,549 @@
...
+                statement_id = response.statement_id
+                _logger.info("Retrying to get statement execution status...")
+                retry_cnt = 1
+                while retry_cnt < 7:
+                    time.sleep(2**retry_cnt)
+                    _logger.info(f"Retry times: {retry_cnt}")
+                    response = self.client.statement_execution.get_statement(statement_id)
+                    if response.status is None or response.status.state != StatementState.PENDING:
+                        break
+                    retry_cnt += 1
```
harupy:
```suggestion
                for _ in range(max_retries):
                    time.sleep(2**retry_cnt)
                    _logger.info(f"Retry times: {retry_cnt}")
                    response = self.client.statement_execution.get_statement(statement_id)
                    if response.status is None or response.status.state != StatementState.PENDING:
                        break
```

Can we use a loop instead?

unitycatalog/functions/databricks.py
```diff
@@ -0,0 +1,549 @@
...
+
+    match = pattern.search(sql_body)
+    if match:
+        return match.group(1)
+    raise ValueError(
+        "Could not extract function name from the sql body. Please "
+        "make sure the sql body follows the syntax of CREATE FUNCTION "
+        "statement in Databricks: "
+        "https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-sql-function.html#syntax."
+    )
```
harupy: can we include `sql_body` in the message?

unitycatalog/functions/client.py
BenWilson2: +1 to this

pyproject.toml
BenWilson2: Let's capitalize `Python` too :) https://devguide.python.org/documentation/style-guide/#specific-words

README.md
```diff
@@ -1 +1,72 @@
 ## Unitycatalog function library
+
```
BenWilson2, line 2: Let's change references to be `Unity Catalog` (similar to Databricks Docs)

README.md
```diff
@@ -1 +1,72 @@
...
+The Unitycatalog function library provides convenient APIs to interact with Unitycatalog functions, including functions creation, retrieval and execution.
+The library includes clients for interacting with both Unitycatalog OSS server and Databricks-managed Unitycatalog.
```
BenWilson2:
```suggestion
The library includes clients for interacting with both the Open-Source Unity Catalog server and the Databricks-managed Unity Catalog service.
```

README.md
```diff
@@ -1 +1,72 @@
...
+pip install git+ssh://git@github.com/serena-ruan/unitycatalog-function.git
+```
+> [!NOTE]
+> Once this package is published to PyPI, users can install via `pip install unitycatalog-function`
+
+
+## Get started
+
+### Databricks-managed UC
+To use Databricks-managed Unitycatalog, please follow the [instruction](https://docs.databricks.com/en/dev-tools/cli/authentication.html#authentication-for-the-databricks-cli) to authenticate first and make sure the token has workspace-level privilege for managing UC functions.
```
BenWilson2:
```suggestion
To use Databricks-managed Unity Catalog with this package, follow the [instructions](https://docs.databricks.com/en/dev-tools/cli/authentication.html#authentication-for-the-databricks-cli) to authenticate to your Workspace and ensure that your access token has workspace-level privilege for managing UC functions.
```

unitycatalog/functions/databricks.py
```diff
@@ -0,0 +1,549 @@
...
+        raise ImportError(
+            "Could not import databricks-sdk python package. "
+            "If you want to use databricks backend then "
+            "please install it with `pip install databricks-sdk`."
+        ) from e
+    return WorkspaceClient()
+
+
+def column_type_to_python_type(column_type: Enum) -> Any:
+    from databricks.sdk.service.catalog import ColumnTypeName
```
BenWilson2: Is there a similar mechanism for the OSS UC catalog to retrieve the type conversion fields? If so, should we externalize an abstract implementation of this into a common util module?

unitycatalog/functions/databricks.py
```diff
@@ -0,0 +1,549 @@
...
+        # it's a type that can be defined in scala, python shouldn't force check the type here
+        # ref: https://www.waitingforcode.com/apache-spark-sql/used-defined-type/read
+        ColumnTypeName.USER_DEFINED_TYPE: object,
+    }
+    if column_type not in mapping:
+        raise ValueError(f"Unsupported column type: {column_type}")
+    return mapping[column_type]
+
+
+def validate_param(param: Any, param_info: "FunctionParameterInfo"):
```
BenWilson2: Similar comment as above. It might be worthwhile to abstract this as a util that both Databricks and UC can use with conditional imports for the `ColumnTypeName` entries that are required for type validation. 
I like the idea of this module having only the client interface APIs in it.

README.md
serena-ruan: Ah actually, databricks-sdk supports executing command against a cluster, it doesn't support serverless. Let me add support of using cluster_id with databricks-sdk in a follow-up PR and then update this section :)

unitycatalog/functions/client.py
serena-ruan: Ah it's because I check if there're leftovers in the parameters on line 72, and raise ValueError if there are.

unitycatalog/functions/client.py
serena-ruan: Yep!

unitycatalog/functions/client.py
serena-ruan: It's not used yet, it will be used in langchainToolkit, something like below when defining the tool's func:
```
def func(*args: Any, **kwargs: Any) -> str:
        client = get_uc_function_client()
        args_json = json.loads(json.dumps(kwargs, default=str))
        result = client.execute_function(
            function_name=func_name,
            parameters=args_json,
        )
        return result.to_json()
```

unitycatalog/functions/databricks.py
serena-ruan: Make sense! OSS UC uses Literal["Boolean", ...] with the same value as this mapping here, this is Enum so we can also get the string value!

unitycatalog/functions/databricks.py
serena-ruan: I'm overriding it just for the docstring, is there any other ways to do it?

src/unitycatalog/functions/client.py
```diff
@@ -0,0 +1,139 @@
...
+    def __init__(self, client: Any = None, **kwargs: Any) -> None:  # noqa: ARG002
+        self._lock = threading.Lock()
+
+    @abstractmethod
+    def create_function(self, function_info: Any) -> Any:
+        """Create a function"""
+
+    @abstractmethod
+    def get_function(self, function_name: str, **kwargs: Any) -> Any:
+        """Get a function by its name"""
```
harupy: can we add docstring for the arguments?

src/unitycatalog/functions/client.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self, client: Any = None, **kwargs: Any) -> None:  # noqa: ARG002
+        self._lock = threading.Lock()
+
+    @abstractmethod
+    def create_function(self, function_info: Any) -> Any:
+        """Create a function"""
+
+    @abstractmethod
+    def get_function(self, function_name: str, **kwargs: Any) -> Any:
```
harupy, line 46: Any constraint on the return type? Can this function return any object?

src/unitycatalog/functions/databricks.py
```diff
@@ -0,0 +1,587 @@
...
+
+class DatabricksFunctionClient(BaseFunctionClient):
+    """
+    Databricks UC function calling client
+    """
+
+    def __init__(self, client: Optional["WorkspaceClient"] = None, **kwargs: Any) -> None:
+        if client is None:
+            client = get_default_databricks_workspace_client()
+        self.client = client
```
harupy:
```suggestion
        self.client = client or get_default_databricks_workspace_client()
```

tests/functions/test_databricks_client.py
```diff
@@ -0,0 +1,729 @@
...
+
+CATALOG = "ml"
+SCHEMA = "serena_uc_test"
+
+_logger = logging.getLogger(__name__)
+
+
+# TODO: CI -- two test cases
+# 1. python 3.10 with databricks-connect 15.1.0, no cluster_id, use serverless
+# 2. python 3.9 with databricks-sdk, with cluster_id
```
harupy, line 39: Even if it's in an early development phase, I'd recommend adding a minimum CI to prevent silly mistakes. We don't need to test everything. Just run the following

1. Install the package
2. Run pytest

unitycatalog/functions/databricks.py
harupy: got it, didn't notice that.

unitycatalog/functions/client.py
harupy: I see, I think I'd use the following:

```python
if keys := parameters.keys() - {p.name for p in input_params.parameters}:
    ...
```

pyproject.toml
```diff
@@ -0,0 +1,72 @@
...
+  "databricks-sdk>=0.32.0",
+  # Note: databricks-connect 15.1 is recommended for serverless, databricks-connect>=14.0.0 requires python>=3.10
+  "databricks-connect==15.1.0",
+  "pandas",
+  "ruff==0.6.4",
+]
+
+[tool.ruff]
+line-length = 100
+target-version = "py38"
```
harupy:
```suggestion
target-version = "py39"
```

src/unitycatalog/functions/client.py
serena-ruan: The return types for UC OSS and Databricks UC are different, and I don't want to depend on these two different packages in this file, so I just set Any

tests/functions/test_databricks_client.py
serena-ruan: Emm it won't work in the CI now, because it needs authentication to a databricks workspace and a warehouse & cluster for testing everything, but I always run the tests locally to make sure they pass before pushing.
image

harupy: Got it. I think you can still test the package installation.

harupy: > it needs authentication to a databricks workspace and a warehouse

Can we use mocks? or you could create a fake Databricks using FastAPI.

harupy: We don't need to solve this right now, but imagine an external contributor files a PR in the future. He/she can't access secrets = he/she can't run tests for Databricks.

Relying on external resources like Databricks has a risk. If it's down, we can't run tests, which slows us down.

serena-ruan: I'll mark those tests involving databricks integration to depend on an environment variable and skip in the normal tests. The rest can be run directly -- mainly for checking the logics.

.github/workflows/main.yml
```diff
@@ -0,0 +1,34 @@
...
+      - master
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+
+jobs:
+    test:
```
harupy: nit: indent

.github/workflows/main.yml
```diff
@@ -0,0 +1,34 @@
...
+        - name: Checkout code
+          uses: actions/checkout@v4
+        - name: Set up Python
+          uses: actions/setup-python@v5
+          with:
+            python-version: ${{ matrix.python-version }}
+        - name: Install dependencies
+          run: |
+            python -m pip install --upgrade pip
+            pip install -r requirements/test-requirements.txt
```
harupy: Is `pip install .` missing?

src/unitycatalog/functions/client.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+
+    @abstractmethod
+    def list_functions(
+        self,
+        catalog: str,
+        schema: str,
+        max_results: Optional[int] = None,
+        page_token: Optional[str] = None,
+    ) -> PagedList[List[Any]]:
```
harupy:
```suggestion
    ) -> PagedList[Any]:
```

harupy (approved):
LGTM

src/unitycatalog/functions/utils.py
```diff
@@ -0,0 +1,109 @@
...
+    Convert a timedelta object to a string representing an interval in the format of 'INTERVAL "d hh:mm:ss.ssssss"'.
+    """
+    days = time_val.days
+    hours, remainder = divmod(time_val.seconds, 3600)
+    minutes, seconds = divmod(remainder, 60)
+    microseconds = time_val.microseconds
+    return f"INTERVAL '{days} {hours}:{minutes}:{seconds}.{microseconds}' DAY TO SECOND"
+
+
+def validate_full_function_name(function_name: str) -> List[str]:
```
harupy:
```suggestion
def validate_full_function_name(function_name: str) -> Tuple[str, str, str]:
```

I'd return a tuple or a named tuple. `List[str]` basically means "not sure how many elements it has."

.github/workflows/test-package-build.yml
```diff
@@ -0,0 +1,68 @@
...
+            echo "sdist-path=${sdist_path}" >> $GITHUB_OUTPUT
+            echo "wheel-path=${wheel_path}" >> $GITHUB_OUTPUT
+            echo "wheel-name=${wheel_name}" >> $GITHUB_OUTPUT
+            echo "wheel-size=${wheel_size}" >> $GITHUB_OUTPUT
+        - name: List files in source distribution
+          run: |
+            tar -tf ${{ steps.build-dist.outputs.sdist-path }}
+        - name: List files in binary distribution
+          run: |
+            unzip -l ${{ steps.build-dist.outputs.wheel-path }}
```
harupy: We can remove these steps. MLflow includes lots of non-python files so inspecting them makes sense, but this package doesn't.

serena-ruan:
Thanks so much for the detailed review @harupy !!! I'll merge this first, and if there're any potential issues I can fix later

PR #4, opened 2024-09-12
Langchain toolkit support

serena-ruan:
Integration with Langchain for supporting UC functions as Langchain tools.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,4 +76,4 @@ jobs:
           pip install integrations/langchain[dev]
       - name: Run tests
         run: |
-          pytest tests
+          pytest integrations/langchain/tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,10 +73,7 @@ jobs:
         run: |
           # for installing unitycatalog-ai
           pip install .
-          cd integrations/langchain
-          python -m pip install --upgrade pip
-          pip install .
-          pip install -r requirements/test-requirements.txt
+          pip install integrations/langchain[dev]
       - name: Run tests
         run: |
           pytest tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,3 +55,29 @@ jobs:
         run: |
           pip install 'pydantic>=2'
           pytest tests/ai/
+
+  langchain_test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.9', '3.10']
+    timeout-minutes: 20
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          # for installing unitycatalog-ai
+          pip install .
+          cd integrations/langchain
+          python -m pip install --upgrade pip
+          pip install .
+          pip install -r requirements/test-requirements.txt
+      - name: Run tests
+        run: |
+          pytest tests
+
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,4 +80,3 @@ jobs:
       - name: Run tests
         run: |
           pytest tests
-
--- a/integrations/langchain/README.md
+++ b/integrations/langchain/README.md
@@ -1,0 +1,109 @@
+## 🦜🔗 Using Unity Catalog AI with Langchain
+
+Integrate Unity Catalog AI package with Langchain to allow seamless usage of UC functions as tools in agents application.
+
+## Installation
+
+```sh
+# install from the source
+pip install git+ssh://git@github.com/serena-ruan/unitycatalog-ai.git#subdirectory=integrations/langchain
+```
+
+> [!NOTE]
+> Once this package is published to PyPI, users can install via `pip install unitycatalog-ai-langchain`
+
+## Get started
+
+### Databricks-managed UC
+
+To use Databricks-managed Unity Catalog with this package, follow the [instructions](https://docs.databricks.com/en/dev-tools/cli/authentication.html#authentication-for-the-databricks-cli) to authenticate to your workspace and ensure that your access token has workspace-level privilege for managing UC functions.
+
+#### Client setup
+Initialize a client for managing UC functions in a Databricks workspace, and set it as the global client.
+
+```python
+from unitycatalog.ai.client import set_uc_function_client
+from unitycatalog.ai.databricks import DatabricksFunctionClient
+
+client = DatabricksFunctionClient(
+    warehouse_id="..." # replace with the warehouse_id
+    cluster_id="..." # optional, only pass when you want to use cluster for function creation
+)
+
+# sets the default uc function client
+set_uc_function_client(client)
+```
+
+#### Create a function in UC
+Create a python UDF in Unity Catalog with the client
+
+```python
+# replace with your own catalog and schema
+CATALOG = "catalog"
+SCHEMA = "schema"
+
+func_name = f"{CATALOG}.{SCHEMA}.python_exec"
+# define the function body in SQL
+sql_body = f"""CREATE OR REPLACE FUNCTION {func_name}(code STRING COMMENT 'Python code to execute. Remember to print the final result to stdout.')
+RETURNS STRING
+LANGUAGE PYTHON
+COMMENT 'Executes Python code and returns its stdout.'
+AS $$
+    import sys
+    from io import StringIO
+    stdout = StringIO()
+    sys.stdout = stdout
+    exec(code)
+    return stdout.getvalue()
+$$
+"""
+
+client.create_function(sql_function_body=sql_body)
+```
+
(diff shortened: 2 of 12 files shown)


integrations/langchain/tests/test_toolkit.py
```diff
@@ -0,0 +1,160 @@
...
+
+        toolkit = LangchainToolkit(function_names=[f"{CATALOG}.{SCHEMA}.*"])
+        assert len(toolkit.tools) >= 1
+        assert func_obj.full_function_name in [t.name for t in toolkit.tools]
+
+
+@pytest.mark.skipif(
+    os.environ.get("TEST_IN_DATABRICKS", "false").lower() != "true",
+    reason="This function test relies on connecting to a databricks workspace",
+)
```
harupy: It looks like this marker is used a lot. Can we create a decorator that looks like:

```python
def requires_databricks(test_func):
    return pytest.mark.skipif(
        os.environ.get("TEST_IN_DATABRICKS", "false").lower() != "true",
        reason="This function test relies on connecting to a databricks workspace",
    )(test_func)
```

integrations/langchain/requirements/test-requirements.txt
```diff
@@ -0,0 +1,8 @@
+unitycatalog-ai==0.0.1
```
harupy: can we install the development version from the source code?

integrations/langchain/README.md
```diff
@@ -0,0 +1,109 @@
...
+from unitycatalog.ai.client import set_uc_function_client
+from unitycatalog.ai.databricks import DatabricksFunctionClient
+
+client = DatabricksFunctionClient(
+    warehouse_id="..." # replace with the warehouse_id
+    cluster_id="..." # optional, only pass when you want to use cluster for function creation
+)
+
+# sets the default uc function client
+set_uc_function_client(client)
```
harupy, line 34: Instead of globally setting the client, can we pass the client to `LangchainToolkit`?

integrations/langchain/pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+authors = [
+    { name="Serena Ruan", email="serena.ruan@databricks.com" }
+]
+readme = "README.md"
+license = { text="MIT" }
+requires-python = ">=3.9"
+dependencies = [
+    "langchain>=0.2.0",
+    "langchain-community>=0.2.0",
+    "unitycatalog-ai==0.0.1",
```
harupy: can we unpin unitycatalog-ai?

integrations/langchain/src/unitycatalog_ai_langchain/toolkit.py
```diff
@@ -0,0 +1,135 @@
...
+        client = validate_or_set_default_client(values.get("client"))
+        values["client"] = client
+
+        function_names = values["function_names"]
+        tools_dict = values["tools_dict"]
+        for name in function_names:
+            if name not in tools_dict:
+                full_func_name = validate_full_function_name(name)
+                if full_func_name.function_name == "*":
+                    token = "None"
```
harupy:
```suggestion
                    token = None
```

why is this a string?

integrations/langchain/src/unitycatalog_ai_langchain/toolkit.py
```diff
@@ -0,0 +1,135 @@
...
+                    while token is not None:
+                        functions = client.list_functions(
+                            catalog=full_func_name.catalog_name,
+                            schema=full_func_name.schema_name,
+                            max_results=os.environ.get(
+                                "UC_LIST_FUNCTIONS_MAX_RESULTS", UC_LIST_FUNCTIONS_MAX_RESULTS
+                            ),
+                            page_token=token if token != "None" else None,
+                        )
```
harupy: what would happen if I had 1M functions in my schema? Would this loop fetch all of them?

integrations/langchain/README.md
serena-ruan: We can, but I think the more general use case is that the user uses the same client across multiple toolkits

integrations/langchain/src/unitycatalog_ai_langchain/toolkit.py
serena-ruan: Yea that's a problem, we can add a `max_functions` limit

integrations/langchain/README.md
harupy: Can we pass the same client to the multiple toolkits?

```python
client = ...
t1 = LangchainToolkit(client=client)
t2 = LangchainToolkit(client=client)
t3 = LangchainToolkit(client=client)
```

harupy: Is there a case that you want to use different clients with different toolkits?

```python
client1 = ...
client2 = ...
t1 = LangchainToolkit(client=client1)
t2 = LangchainToolkit(client=client2)
```

integrations/langchain/src/unitycatalog_ai_langchain/toolkit.py
harupy: why do we need to fetch all of them? Can we just fetch the required functions?

.github/workflows/main.yml
```diff
@@ -55,3 +55,28 @@ jobs:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          # for installing unitycatalog-ai
+          pip install .
+          cd integrations/langchain
+          python -m pip install --upgrade pip
+          pip install .
```
harupy:
```suggestion
          python -m pip install --upgrade pip
          pip install integrations/langchain
```

.github/workflows/main.yml
```diff
@@ -55,3 +55,28 @@ jobs:
...
+          pip install -r requirements/test-requirements.txt
```
harupy: can we install test dependencies using an extra?

integrations/langchain/tests/test_toolkit.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+@pytest.fixture
+def client() -> DatabricksFunctionClient:
+    with mock.patch(
+        "unitycatalog.ai.databricks.get_default_databricks_workspace_client",
+        return_value=mock.Mock(),
+    ):
+        return DatabricksFunctionClient(warehouse_id="warehouse_id", cluster_id="cluster_id")
```
harupy, line 29:
```suggestion
        yield DatabricksFunctionClient(warehouse_id="warehouse_id", cluster_id="cluster_id")
```

integrations/langchain/README.md
serena-ruan: Yes we can pass the same client to multiple toolkits; The second case is also possible, pls see https://docs.google.com/document/d/1pNmXvkAt8x70XuvfFJESXlbRN2OA19SWFfW1QU7yokQ/edit?pli=1#bookmark=id.qbel1z2ohit7 for this decision :)

integrations/langchain/src/unitycatalog_ai_langchain/toolkit.py
serena-ruan: So that I can enter the while loop `while token is not None`

integrations/langchain/src/unitycatalog_ai_langchain/toolkit.py
serena-ruan: Emm what do you mean by `required functions`? If users use `catalog.schema.*` that means the user wants to include all functions in that schema; otherwise the user could specify the function name for a specific function

harupy: For example, if `function_names` is `["foo", "bar"]`, we only need `foo` and `bar`.

integrations/langchain/src/unitycatalog_ai_langchain/toolkit.py
harupy: Let's modify the code. `token = "None"` does look like an error/a mistake.

integrations/langchain/src/unitycatalog_ai_langchain/toolkit.py
harupy: > If users use catalog.schema.* that means the user wants to include all functions in that schema

Got it :)

integrations/langchain/src/unitycatalog_ai_langchain/toolkit.py
```diff
@@ -0,0 +1,142 @@
...
+                full_func_name = validate_full_function_name(name)
+                if full_func_name.function_name == "*":
+                    token = "None"
+                    while token is not None:
+                        functions = client.list_functions(
+                            catalog=full_func_name.catalog_name,
+                            schema=full_func_name.schema_name,
+                            max_results=os.environ.get(
+                                "UC_LIST_FUNCTIONS_MAX_RESULTS", UC_LIST_FUNCTIONS_MAX_RESULTS
+                            ),
```
harupy: `os.environ.get` returns a string, not an integer if `UC_LIST_FUNCTIONS_MAX_RESULTS` is exists. `os.environ.get("foo", non_string_object)` is error-prone.

harupy (approved):
LGTM

PR #6, opened 2024-09-19
Add client execution timeout and add test

serena-ruan:
Add CLIENT_EXECUTION_TIMEOUT for controlling the client side wait time for executing functions.
Note: the default wait_timeout in SQL execution side is 30s, so the wait time on the execute_statement is controlled by wait_timeout parameter; while CLIENT_EXECUTION_TIMEOUT controls how long we should keep retrying for getting a response.
TODO: we should add an example explaining this in a follow-up PR.

### DIFF
--- a/src/unitycatalog/ai/databricks.py
+++ b/src/unitycatalog/ai/databricks.py
@@ -37,6 +37,7 @@ DEFAULT_EXECUTE_FUNCTION_ARGS = {
     "row_limit": 100,
     "byte_limit": 4096,
 }
+UNITICATALOG_AI_CLIENT_EXECUTION_TIMEOUT = "UNITICATALOG_AI_CLIENT_EXECUTION_TIMEOUT"
 
 _logger = logging.getLogger(__name__)
 
@@ -410,7 +411,9 @@ class DatabricksFunctionClient(BaseFunctionClient):
                 _logger.info("Retrying to get statement execution status...")
                 wait_time = 0
                 retry_cnt = 0
-                client_execution_timeout = int(os.environ.get("CLIENT_EXECUTION_TIMEOUT", "120"))
+                client_execution_timeout = int(
+                    os.environ.get(UNITICATALOG_AI_CLIENT_EXECUTION_TIMEOUT, "120")
+                )
                 while wait_time < client_execution_timeout:
                     wait = min(2**retry_cnt, client_execution_timeout - wait_time)
                     time.sleep(wait)
@@ -424,7 +427,7 @@ class DatabricksFunctionClient(BaseFunctionClient):
                     return FunctionExecutionResult(
                         error=f"Statement execution is still pending after {wait_time} "
                         "seconds. Please try increase the wait_timeout argument for executing "
-                        "the function or increase CLIENT_EXECUTION_TIMEOUT environment "
+                        f"the function or increase {UNITICATALOG_AI_CLIENT_EXECUTION_TIMEOUT} environment "
                         "variable for increasing retrying time."
                     )
             if response.status is None:
--- a/src/unitycatalog/ai/databricks.py
+++ b/src/unitycatalog/ai/databricks.py
@@ -38,6 +38,7 @@ DEFAULT_EXECUTE_FUNCTION_ARGS = {
     "byte_limit": 4096,
 }
 UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT = "UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT"
+DEFAULT_UC_AI_CLIENT_EXECUTION_TIMEOUT = "120"
 
 _logger = logging.getLogger(__name__)
 
@@ -412,7 +413,10 @@ class DatabricksFunctionClient(BaseFunctionClient):
                 wait_time = 0
                 retry_cnt = 0
                 client_execution_timeout = int(
-                    os.environ.get(UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT, "120")
+                    os.environ.get(
+                        UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT,
+                        DEFAULT_UC_AI_CLIENT_EXECUTION_TIMEOUT,
+                    )
                 )
                 while wait_time < client_execution_timeout:
                     wait = min(2**retry_cnt, client_execution_timeout - wait_time)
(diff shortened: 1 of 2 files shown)


src/unitycatalog/ai/databricks.py
```diff
@@ -408,17 +408,24 @@ def _execute_uc_function(
             if response.status and job_pending(response.status.state) and response.statement_id:
                 statement_id = response.statement_id
                 _logger.info("Retrying to get statement execution status...")
-                max_retries = 6
-                for retry_cnt in range(1, max_retries + 1):
-                    time.sleep(2**retry_cnt)
+                wait_time = 0
+                retry_cnt = 0
+                client_execution_timeout = int(os.environ.get("CLIENT_EXECUTION_TIMEOUT", "120"))
```
```suggestion
                client_execution_timeout = int(os.environ.get("UNITICATALOG_AI_CLIENT_EXECUTION_TIMEOUT", "120"))
```

Can we add a prefix to avoid conflicts with other packages?

src/unitycatalog/ai/databricks.py
```diff
@@ -408,17 +409,26 @@ def _execute_uc_function(
...
                     if response.status is None or not job_pending(response.status.state):
                         break
+                    wait_time += wait
+                    retry_cnt += 1
                 if response.status and job_pending(response.status.state):
                     return FunctionExecutionResult(
-                        error=f"Statement execution is still pending after {max_retries} times. "
-                        "Please try increase the wait_timeout argument for executing the function."
+                        error=f"Statement execution is still pending after {wait_time} "
+                        "seconds. Please try increase the wait_timeout argument for executing "
```
```suggestion
                        "seconds. Please increase the wait_timeout argument for executing "
```

src/unitycatalog/ai/databricks.py
```diff
@@ -408,17 +409,26 @@ def _execute_uc_function(
...
+                        f"the function or increase {UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT} environment "
+                        "variable for increasing retrying time."
```
harupy: I think we should mention the default value.

src/unitycatalog/ai/databricks.py
```diff
@@ -1,42 +1,44 @@
...
+    from databricks.sdk.service.sql import StatementParameterListItem, StatementState
 
 EXECUTE_FUNCTION_ARG_NAME = "__execution_args__"
 DEFAULT_EXECUTE_FUNCTION_ARGS = {
     "wait_timeout": "30s",
     "row_limit": 100,
     "byte_limit": 4096,
 }
+UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT = "UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT"
+DEFAULT_UC_AI_CLIENT_EXECUTION_TIMEOUT = "120"
```
harupy, line 41: Not a blocker, but I'd recommend put these environment variables in one place like mlflow.

tests/ai/test_databricks_client.py
```diff
@@ -262,6 +264,37 @@ def test_create_and_execute_function(
...
+        assert result.error.startswith("Statement execution is still pending after 5 seconds")
+
+        time2 = time.time()
+        result = client.execute_function(
+            full_func_name, {EXECUTE_FUNCTION_ARG_NAME: {"wait_timeout": "10s"}}
+        )
+        time_total2 = time.time() - time2
+        assert result.error.startswith("Statement execution is still pending after 5 seconds")
+        # default wait_timeout is 30s
+        assert abs(abs(time_total2 - time_total1) - 20) < 5
```
harupy: where does 20 come from?

serena-ruan: default wait_timeout in `DEFAULT_EXECUTE_FUNCTION_ARGS` passed to `self.client.statement_execution.execute_statement` is 30s, and the second test we override that to 10s, 30-10=20

harupy: do we need this line?

serena-ruan: Yea I want to make sure the parameters passed as EXECUTE_FUNCTION_ARG_NAME works, as part of the integration test. But I can split it to a different test if you think this is too much?

src/unitycatalog/ai/databricks.py
serena-ruan: I'll put this inside this file for now, just in case there're other env vars coming from OSS integration. If that's true, we'd better add a prefix if unifying them in one place? Wondering if that would be quite a long name

tests/ai/test_databricks_client.py
harupy: Let's split it up.

serena-ruan:
Porting over the logic here to langchain: https://github.com/langchain-ai/langchain/pull/26645 for supporting agent team's work

tests/ai/test_databricks_client.py
```diff
@@ -739,3 +766,33 @@ def mock_execute_statement(
...
+
+    time.sleep(100)
+    return "10"
+$$
+"""
+        client.create_function(sql_function_body=sql_body)
+        time1 = time.time()
+        # default wait_timeout is 30s
+        client.execute_function(full_func_name)
+        time_total1 = time.time() - time1
```
harupy, line 789:
```suggestion
        time_total1 = time.time() - time1
        assert 29 <= time_total1 <= 31
```

Is this more straightforward?

src/unitycatalog/ai/databricks.py
```diff
@@ -408,17 +410,29 @@ def _execute_uc_function(
...
+                wait_time = 0
+                retry_cnt = 0
+                client_execution_timeout = int(
+                    os.environ.get(
+                        UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT,
+                        DEFAULT_UC_AI_CLIENT_EXECUTION_TIMEOUT,
+                    )
+                )
+                while wait_time < client_execution_timeout:
+                    wait = min(2**retry_cnt, client_execution_timeout - wait_time)
```
harupy, line 422: any chance that `client_execution_timeout - wait_time` becomes negative?

harupy: never mind, it's always positive

serena-ruan: The loop requirement is wait_time < client_execution_timeout

tests/ai/test_databricks_client.py
serena-ruan: This is actually not true, because the total execution time include wait_timeout + retry time + data parsing time

harupy: I'm just not big fan of `assert abs(abs(time_total2 - time_total1) - 20) < 5`. I'm fine if that's not flaky.

harupy (approved):
LGTM

tests/ai/test_databricks_client.py
serena-ruan: I think 5 seconds is enough buffer for 'flaky' tolerance :) We'll see how it goes!

PR #7, opened 2024-09-19
Support serverless execution

serena-ruan:
Support execution using serverless compute in databricks UC.

Test validated locally:
image

Currently only row_limit is supported, as limiting byte_size on the spark dataframe result is non-trivial, we can add support later if there're certain requests.
The sql_command generation function is similar to `get_execute_function_sql_stmt` function, but it produces a single sql string so I didn't unify them, otherwise it might get quite messy.

Limitation:
Ideally we should support passing parameters into `spark.sql(sqlQuery=..., args=..., kwargs=...)` to avoid code injection, but with my test it currently doesn't work:
image

image

### DIFF
--- a/src/unitycatalog/ai/databricks.py
+++ b/src/unitycatalog/ai/databricks.py
@@ -39,6 +39,9 @@ DEFAULT_EXECUTE_FUNCTION_ARGS = {
 }
 UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT = "UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT"
 DEFAULT_UC_AI_CLIENT_EXECUTION_TIMEOUT = "120"
+UC_AI_CLIENT_EXECUTION_RESULT_ROW_LIMIT = "UC_AI_CLIENT_EXECUTION_RESULT_ROW_LIMIT"
+DEFAULT_UC_AI_CLIENT_EXECUTION_RESULT_ROW_LIMIT = "100"
+
 
 _logger = logging.getLogger(__name__)
 
@@ -396,6 +399,7 @@ class DatabricksFunctionClient(BaseFunctionClient):
                 )
 
         if self.warehouse_id:
+            _logger.info("Executing function using client warehouse_id.")
             parametrized_statement = get_execute_function_sql_stmt(function_info, parameters)
             response = self.client.statement_execution.execute_statement(
                 statement=parametrized_statement.statement,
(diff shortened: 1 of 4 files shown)


tests/ai/test_databricks_client.py
```diff
@@ -48,6 +49,11 @@ def client() -> DatabricksFunctionClient:
         return DatabricksFunctionClient(warehouse_id="warehouse_id", cluster_id="cluster_id")
 
 
+@pytest.fixture
+def client_using_serverless() -> DatabricksFunctionClient:
```
harupy:
```suggestion
def serverless_client() -> DatabricksFunctionClient:
```

src/unitycatalog/ai/databricks.py
```diff
@@ -620,3 +652,65 @@ def get_execute_function_sql_stmt(
...
+                    )
+                    arg_clause += interval_str
+                else:
+                    if param_info.type_name == ColumnTypeName.DECIMAL and isinstance(
+                        param_value, Decimal
+                    ):
+                        _logger.warning(
+                            f"Param {param_info.name} has decimal value {param_value}, it is converted to float "
+                            "for execution, please note that this conversion may lose precision."
+                        )
```
BenWilson2: Curious if we need this or not. This would theoretically be hit on every request for a given agent if it invokes a tool that is querying a table that contains this type... but builders of the app would be aware of this limitation. With the Decimal -> float casting, could we just leave this as a warning in the docs?

tests/ai/test_databricks_client.py
```diff
@@ -93,17 +99,17 @@ def function_with_struct_input(func_name: str) -> FunctionInputOutput:
...
-    return [str(i) for i in s]
+    return ",".join(str(i) for i in s)
 $$
 """
     return FunctionInputOutput(
         sql_body=sql_body,
         func_name=f"{CATALOG}.{SCHEMA}.{func_name}",
         inputs=[{"s": [1, 2, 3]}],
-        output='["1","2","3"]',
+        output="1,2,3",
```
BenWilson2, line 112: Much nicer :)

BenWilson2:
Could we verify what the handling will be like for the csv dump for a multi-column, multi-row response just to make sure that we're covering the handling parsing properly (a mocked return test - just to be safe :) )

BenWilson2 (approved):
Overall LGTM - let's get that test added just to protect us from some potential odd behavior when returning table results :)

serena-ruan:
> let's get that test added just to protect us from some potential odd behavior when returning table results :)

Actually function_with_table_output covers that, and we use the serverless client for testing :)

src/unitycatalog/ai/databricks.py
serena-ruan: Make sense!

src/unitycatalog/ai/databricks.py
```diff
@@ -484,8 +488,36 @@ def _execute_uc_function(
...
+                    )
+                    truncated = False
+                    if result.count() > row_limit:
+                        truncated = True
+                    pdf = result.limit(row_limit).toPandas()
+                    csv_buffer = StringIO()
+                    pdf.to_csv(csv_buffer, index=False)
+                    return FunctionExecutionResult(
+                        format="CSV", value=csv_buffer.getvalue(), truncated=truncated
+                    )
```
harupy: Is it possible minimize the try-catch scope?

src/unitycatalog/ai/databricks.py
```diff
@@ -484,8 +488,36 @@ def _execute_uc_function(
...
+                else:
+                    row_limit = int(
+                        os.environ.get(
+                            UC_AI_CLIENT_EXECUTION_RESULT_ROW_LIMIT,
+                            DEFAULT_UC_AI_CLIENT_EXECUTION_RESULT_ROW_LIMIT,
+                        )
+                    )
+                    truncated = False
+                    if result.count() > row_limit:
+                        truncated = True
```
```suggestion
                    truncated = result.count() > row_limit
```

src/unitycatalog/ai/databricks.py
serena-ruan: Emm I moved first two lines out. 
I think `result = self.spark.sql(sqlQuery=sql_command)` needs to be in the try catch, so are `result.collect()` and `result.toPandas()` as that's when we do the computation. But still it's quite long code to wrap 🤔

serena-ruan:
Tests validation
image

harupy (approved):
LGTM!

PR #9, opened 2024-09-24
Core utils update

serena-ruan:
* extract some functions to main utils file
* Add `strict` param for pydantic type extraction method in preparation for openAI support
`strict` is True when all fields are JSON schema types
* Fix a bug in langchain toolkit when passing client

Tests validation:
image
image

### DIFF
--- a/src/unitycatalog/ai/utils.py
+++ b/src/unitycatalog/ai/utils.py
@@ -1,13 +1,35 @@
-import datetime
 import base64
+import datetime
 import decimal
 import json
+import logging
 from dataclasses import dataclass
 from hashlib import md5
 from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Type, Union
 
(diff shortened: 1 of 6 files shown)


harupy (approved):
LGTM

src/unitycatalog/ai/utils.py
```diff
@@ -233,24 +280,27 @@ class PydanticField:
     pydantic_type: Type
     description: Optional[str] = None
     default: Optional[Any] = None
+    strict: bool = False
 
 
-def param_info_to_pydantic_type(param_info: Any) -> PydanticField:
+def param_info_to_pydantic_type(param_info: Any, strict=False) -> PydanticField:
```
harupy:
```suggestion
def param_info_to_pydantic_type(param_info: Any, strict: bool=False) -> PydanticField:
```

PR #10, opened 2024-09-24
Openai Toolkit support

serena-ruan:
Add support for OpenAI tool integration
Note: this PR depends on https://github.com/serena-ruan/unitycatalog-ai/pull/9 for some utils update. ~Will rebase once the previous PR is merged.~
Rebased.
Tests are validated locally.


integrations/openai/src/unitycatalog_ai_openai/toolkit.py
```diff
@@ -0,0 +1,89 @@
+from typing import Any, Dict, List, Optional
+
+from openai.lib._tools import pydantic_function_tool
```
harupy: Do we need to use this? This makes the openai integration fragile. Private modules/functions in a third-party modules should not be touched.

integrations/openai/tests/test_toolkit.py
```diff
@@ -0,0 +1,462 @@
...
+        else:
+            return DatabricksFunctionClient(warehouse_id="warehouse_id", cluster_id="cluster_id")
+
+
+def random_func_name():
+    return f"{CATALOG}.{SCHEMA}.test_{uuid.uuid4().hex[:4]}"
+
+
+@contextmanager
+def create_function_and_cleanup(
```
harupy, line 47: looks like each integration has this function. Can we create common utils and import this function from there?

harupy: The same comments to other helper functions.

integrations/openai/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+To use different clients during toolkit creation stage, you could pass the client directly to OpenAIToolkit:
+
+```python
+from unitycatalog_ai_openai.toolkit import OpenAIToolkit
+
+toolkit = OpenAIToolkit(function_names=[...], client=your_own_client)
+```
+
+Please note that this client is only used for retrieving UC functions so we can generate OpenAI accepted function definitions, which you could pass to the OpenAI API call. After getting a response from the OpenAI API, you should be responsible for executing them using the corresponding client with `client.execute_function(...)` as above example.
```
harupy, line 156: can we replace this with code? I think that's more helpful.

integrations/openai/src/ucai_openai/toolkit.py
```diff
@@ -0,0 +1,89 @@
...
+from unitycatalog.ai.client import BaseFunctionClient
+from unitycatalog.ai.utils.client_utils import validate_or_set_default_client
+from unitycatalog.ai.utils.function_processing_utils import (
+    generate_function_input_params_schema,
+    get_tool_name,
+    process_function_names,
+)
+
+
+class UCFunctionToolkit(BaseModel):
```
harupy, line 15: Does this need to be a pydantic model?

integrations/openai/tests/test_toolkit.py
serena-ruan: Then we need to pack them when packing unitycatalog-ai package?

integrations/openai/README.md
serena-ruan: The example above explains how to handle the response, this is just a reminder I think

integrations/openai/src/ucai_openai/toolkit.py
serena-ruan: It doesn't have to be, it's just easier for type validation. Do you have any concerns?

integrations/openai/src/unitycatalog_ai_openai/toolkit.py
serena-ruan: This is actually exposed at openai level, let me change the import :)

integrations/openai/src/ucai_openai/toolkit.py
harupy: Not a big concern, but breaking changes in pydantic.

serena-ruan: I think we can pin pydantic<3 for now :)

integrations/openai/src/ucai_openai/toolkit.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self.tools_dict = process_function_names(
+            function_names=self.function_names,
+            tools_dict={},
+            client=self.client,
+            uc_function_to_tool_func=self.uc_function_to_openai_function_definition,
+        )
+        return self
+
+    @classmethod
```
harupy: we can use a static method. I'd use a class method for a factory such as `from_dict`.

integrations/openai/src/ucai_openai/toolkit.py
harupy: Fine to use pydantic for now. Just wondering what the requirements of a toolkit are.

serena-ruan: I think a normal python class is fine. But the schema extraction process still relies on pydantic, so we're not introducing pydantic as an extra dependency here.

integrations/openai/src/ucai_openai/toolkit.py
serena-ruan: Make sense

harupy (approved):
LGTM!

src/unitycatalog/ai/utils/function_processing_utils.py
```diff
@@ -110,6 +110,26 @@ def get_tool_name(func_name: str) -> str:
     return tool_name
 
 
+def construct_original_function_name(tool_name: str) -> str:
+    """
+    Args:
+        tool_name: The tool name in the form of `catalog__schema__function`.
```
harupy, line 116: Can `catalog` contain `__`?

serena-ruan: I think it can (https://docs.databricks.com/en/sql/language-manual/sql-ref-names.html#catalog-name), but I don't think we have a better way...
We do warn if the total name length is larger than 64. But we shouldn't prevent users from using the functions if the catalog/schema/function name contains `__`, because this function is only used in a helper function, and users can always customize based their own cases.

harupy: what would happen if my function name is `foo.bar.my__func` and I use it as a tool?

serena-ruan: It's totally fine, but if you use `generate_tool_call_messages` helper function it would complain, as the converted tool name in OpenAI would be `foo__bar__my__func` (openAI tool doesn't allow `.` in the tool name), and we try to convert it back by splitting over `__` then it fails. While if you know that the original function name is `foo.bar.my__func` you don't need to rely on `generate_tool_call_messages`, and you can directly execute the function on your own

integrations/openai/tests/test_toolkit.py
harupy: what do you mean by `packing`?

serena-ruan: I mean `building the unitycatalog-ai package`.
If we define this at a place that all sub-pacakges have access to, the only common dependency is `unitycatalog-ai` package.

harupy: can we create a package (e.g. test_utils) in the repository root?

serena-ruan: You mean under tests/test_utils? Let me try this in a follow-up PR :)

PR #13, opened 2024-09-25
Update DatabricksFunctionClient function signature

serena-ruan:
Update init method to be more informative and add docstring;
Make `create_function` only accept kwargs to avoid misuse.

### DIFF
--- a/src/unitycatalog/ai/databricks.py
+++ b/src/unitycatalog/ai/databricks.py
@@ -177,6 +177,7 @@ class DatabricksFunctionClient(BaseFunctionClient):
     @override
     def create_function(
         self,
+        *,
         sql_function_body: Optional[str] = None,
         function_info: Optional["CreateFunction"] = None,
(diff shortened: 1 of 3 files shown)


src/unitycatalog/ai/databricks.py
```diff
@@ -106,11 +106,32 @@ class DatabricksFunctionClient(BaseFunctionClient):
...
+        profile: Optional[str] = None,
+        **kwargs: Any,
+    ) -> None:
+        """
+        Databricks UC function calling client.
+
+        Args:
+            client: The databricks workspace client. If it's None, a default databricks workspace client
+                is generated based on the configuration. Defaults to None.
+            warehouse_id (str, optional): The warehouse id to use for executing functions. This field is
```
```suggestion
            warehouse_id: The warehouse id to use for executing functions. This field is
```

because we have type hints.

harupy (approved):
LGTM!

PR #16, opened 2024-09-26
Refactor tests

serena-ruan:


harupy (approved):
LGTM

src/unitycatalog/test_utils/__init__.py
```diff
@@ -0,0 +1,2 @@
+from unitycatalog.test_utils.client_utils import *
+from unitycatalog.test_utils.function_utils import *
```
harupy: can we avoid `*` import?

tests/ai/test_databricks_client.py
```diff
@@ -30,55 +26,25 @@
...
-        yield func_name
-    finally:
-        try:
-            client.client.functions.delete(f"{CATALOG}.{SCHEMA}.{func_name}")
-        except Exception as e:
-            _logger.warning(f"Fail to delete function: {e}")
+from unitycatalog.test_utils import (
+    CATALOG,
+    SCHEMA,
+    client,  # noqa: F401
```
harupy: why do we need F401?

BenWilson2 (approved):
Looks great! I'll file a PR tomorrow to clean up the Llama tests after this merges!

src/unitycatalog/test_utils/__init__.py
harupy: I'd consider using `*` if I need to import 1000 items.

src/unitycatalog/test_utils/__init__.py
harupy, line 1: can we move `test_utils` in `tests` in the root?

serena-ruan: Emm we can't, it wouldn't be discovered by the integration tests folder

tests/ai/test_databricks_client.py
serena-ruan: Because it's used as fixture in the test but ruff complained 
image

src/unitycatalog/test_utils/__init__.py
serena-ruan: That's true, but what's the concern of using `*`?

harupy: https://stackoverflow.com/a/2386740 explains why

src/unitycatalog/test_utils/__init__.py
harupy: can we install it as a package?

src/unitycatalog/test_utils/__init__.py
harupy: what if client_utils and function_utils have the same function name?

serena-ruan: Make sense :) Updated

src/unitycatalog/test_utils/function_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    return stdout.getvalue()
+$$
+"""
+    )
+    try:
+        client.create_function(sql_function_body=sql_body)
+        if return_func_name:
+            yield func_name
+        else:
+            yield FunctionObj(full_function_name=func_name, comment=comment)
```
harupy: why not always return `FunctionObj`?

PR #18, opened 2024-09-27
Drop cluster_id and refactor

serena-ruan:
After offline discussion, we'd like to drop support of cluster_id when creating function -- it requires the cluster to be ready to run, otherwise the cluster starting process takes a long time and the user experience would be so bad.
Refactor and update:
 - If databricks-connect version doesn't support serverless, directly raise error. Python>=3.10 will be a hard requirement to use databricks client. (DBR >= 13.3 supports python>=3.10, we should be fine)
 - Only serverless warehouse is supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,22 +21,23 @@ To use Databricks-managed Unity Catalog with this package, follow the [instructi
 
 #### Prerequisites
 
+- **[Highly recommended]** Use python>=3.10 for accessing all functionalities including function creation and function execution.
 - Install databricks-sdk package with `pip install databricks-sdk`.
-- For accessing the UC functions in Databricks, please create a SQL warehouse following [this instruction](https://docs.databricks.com/en/compute/sql-warehouse/create.html), and save the warehouse id.
-- [Optional] Install databricks-connect package with `pip install databricks-connect`.
-  - This package is only required for creating UC functions using sql body.
-  - If you want to use [serverless compute](https://docs.databricks.com/en/compute/use-compute.html#use-serverless-compute) for function creation, please make sure you have python>=3.10 and databricks-connect==15.1.0.
-  - If you want to use [your own cluster](https://docs.databricks.com/en/compute/use-compute.html#create-new-compute-using-a-policy) for function creation, please make sure the cluster is up running and pass the cluster_id when creating DatabricksFunctionClient.
+- For creating UC functions with SQL body, **only [serverless compute](https://docs.databricks.com/en/compute/use-compute.html#use-serverless-compute) is supported**.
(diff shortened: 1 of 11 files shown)


serena-ruan:
tests
image
image
image

src/ucai/core/databricks.py
```diff
@@ -382,10 +340,22 @@ def execute_function(
     def _execute_uc_function(
         self, function_info: "FunctionInfo", parameters: Dict[str, Any], **kwargs: Any
     ) -> Any:
+        if self.warehouse_id:
```
serena-ruan, line 343: This change is just splitting into two functions

README.md
```diff
@@ -21,22 +21,23 @@ To use Databricks-managed Unity Catalog with this package, follow the [instructi
 
 #### Prerequisites
 
+- **[Highly recommended]** Use python>=3.10 for accessing all functionalities including function creation and function execution.
```
harupy, line 24: why?

integrations/llama_index/tests/test_toolkit.py
```diff
@@ -114,7 +51,9 @@ def test_toolkit_e2e(use_serverless, monkeypatch):
 
         toolkit = UCFunctionToolkit(function_names=[f"{CATALOG}.{SCHEMA}.*"])
         assert len(toolkit.tools) >= 1
-        assert func_obj.full_function_name in [t.metadata.name for t in toolkit.tools]
+        assert get_tool_name(func_obj.full_function_name) in [
```
harupy: can we add a tool_name field to `FunctionObject`?

src/ucai/core/databricks.py
```diff
@@ -419,124 +388,122 @@ def _execute_uc_function(
...
+                time.sleep(wait)
+                _logger.info(f"Retry times: {retry_cnt}")
+                response = self.client.statement_execution.get_statement(statement_id)
+                if response.status is None or not job_pending(response.status.state):
+                    break
+                wait_time += wait
+                retry_cnt += 1
+            if response.status and job_pending(response.status.state):
+                return FunctionExecutionResult(
+                    error=f"Statement execution is still pending after {wait_time} "
```
```suggestion
                    error=f"Statement execution is still {response.status.state} after {wait_time} "
```

can we include the status?

README.md
serena-ruan: Because databricks-sdk==15.1.0 only supports python>=3.10, and we want customers to have a great journey the first time trying out this client, then `serverless` is highly recommended. And to use function creation feature, the user has to use serverless with dbconnect.
I checked DBR 13.3 already supports python>=3.10, and this is only databricks-customers facing, so I think it's fine.

harupy: got it, thanks for the explanation 👍

harupy (approved):
LGTM

PR #19, opened 2024-09-27
Add NoOp client

serena-ruan:
Add NoOpClient if WorkspaceClient initialization fails.
This is to prevent the client from failing inside subprocess, though it doesn't fixes the case when serverless client is used instead of SQL warehouse.

### DIFF
--- a/src/ucai/core/utils/databricks_utils.py
+++ b/src/ucai/core/utils/databricks_utils.py
@@ -1,0 +1,62 @@
+import logging
+from typing import TYPE_CHECKING
+
+_logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:
+    from databricks.sdk import WorkspaceClient
+
+
+def get_default_databricks_workspace_client() -> "WorkspaceClient":
+    try:
+        from databricks.sdk import WorkspaceClient
+
+        return WorkspaceClient()
+    except ImportError as e:
+        raise ImportError(
+            "Could not import databricks-sdk python package. "
+            "If you want to use databricks backend then "
+            "please install it with `pip install databricks-sdk`."
+        ) from e
+    except Exception as e:
+        _logger.warning(
+            "Failed to initialize databricks workspace client "
+            f"Falling back to NoOpClient. Error: {e}"
+        )
+    return NoOpClient()
+
+
+class NoOpStatementExecutionAPI:
+    def execute_statement(self, *args, **kwargs):
+        from databricks.sdk.service.sql import (
+            ResultData,
+            ResultManifest,
+            StatementResponse,
+            StatementState,
+            StatementStatus,
+        )
+
+        return StatementResponse(
+            manifest=ResultManifest(truncated=False),
+            result=ResultData(data_array=[["null"]]),
+            status=StatementStatus(state=StatementState.SUCCEEDED),
+        )
+
+
+class NoOpAPIClient:
+    def do(self, *args, **kwargs):
+        return {}
+
+
+class NoOpFunctionsAPI:
+    _api: NoOpAPIClient = NoOpAPIClient()
+
+    def get(self, *args, **kwargs):
+        from databricks.sdk.service.catalog import FunctionInfo
+
+        return FunctionInfo()
+
+
+class NoOpClient:
+    statement_execution: NoOpStatementExecutionAPI = NoOpStatementExecutionAPI()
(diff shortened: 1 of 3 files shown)


tests/core/test_databricks_client.py
```diff
@@ -811,3 +812,16 @@ def test_extra_params_when_executing_function_e2e(client: DatabricksFunctionClie
...
+def test_default_noop_client_if_databricks_sdk_missing():
+    with mock.patch("databricks.sdk.WorkspaceClient.__init__", side_effect=RuntimeError):
+        client = DatabricksFunctionClient(warehouse_id="fake_warehouse_id")
+        assert isinstance(client.client, NoOpClient)
+        # create_function is not tested because it doesn't rely on the client
+        function_info = client.get_function("catalog.schema.test")
+        assert function_info is not None
+        assert all(x is None for x in function_info.__dict__.values())
+        assert client.list_functions(catalog="catalog", schema="schema") == []
+        with mock.patch.object(client, "validate_input_params"):
```
serena-ruan: Seems like we need to pass over this check if the client is no-op

harupy:
can we make the client work in a subprocess?

serena-ruan:
> can we make the client work in a subprocess?

There's a ticket to track this, but it's a bug in dbconnect or databricks-sdk, we have no control.

harupy (approved):
LGTM

src/ucai/core/utils/databricks_utils.py
```diff
@@ -0,0 +1,72 @@
...
+            ResultData,
+            ResultManifest,
+            StatementResponse,
+            StatementState,
+            StatementStatus,
+        )
+
+        return StatementResponse(
+            manifest=ResultManifest(truncated=False),
+            result=ResultData(data_array=[["null"]]),
```
harupy: can we pass an empty array to data_array?

src/ucai/core/utils/databricks_utils.py
```diff
@@ -0,0 +1,72 @@
...
+            status=StatementStatus(state=StatementState.SUCCEEDED),
+        )
+
+
+class NoOpAPIClient:
```
harupy: let's insert a comment. It's unclear when to use this class.

src/ucai/core/utils/databricks_utils.py
serena-ruan: We can't as we need to find this field in execute_uc_function, and data_array is of List[List[str]] type.

harupy: what about `[[]]`?

src/ucai/core/utils/databricks_utils.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class _NoOpWarehousesAPI:
+    def get(self, *args, **kwargs):
+        return _NoOpWarehouse()
+
+
+# This is a no-op client that can be used as a fallback when the databricks
+# WorkspaceClient fails to initialize. It mocks the behavior of WorkspaceClient
+# and its methods used in the UCAI codebase, but actually does nothing.
```
harupy, line 71: Can we define a wrapper class that wraps the client, and define required methods in that class. If the real databricks client is unavailable, those methods return dummy values.

src/ucai/core/utils/databricks_utils.py
serena-ruan: Emm we use `data_array[0][0]` to fetch the value, if the value is None, then normally the returned data_array would be `[["null"]]` as it's json serialized

src/ucai/core/utils/databricks_utils.py
serena-ruan: I don't want to wrap the whole WorkspaceClient as it has so many methods. Also I think once dbconnect resolves their issue we can remove this client.

src/ucai/core/utils/databricks_utils.py
harupy: looks like we have a guard `data_array and len(data_array) > 0 and len(data_array[0]) > 0`:

https://github.com/serena-ruan/unitycatalog-ai/blob/ed7c0ba39be89b51e91bd8105115ecbaf2d5d030/src/ucai/core/databricks.py#L451-L455

serena-ruan:
Close this PR as we found workaround for subprocess connecting workspaceClient.

PR #20, opened 2024-09-30
Update tests for integration tests

serena-ruan:


serena-ruan:
This PR is verified by the integration test workflow.

integrations/langchain/pyproject.toml
```diff
@@ -15,6 +15,14 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+databricks-dev = [
```
harupy: where is this used?

src/ucai/core/databricks.py
```diff
@@ -42,14 +42,18 @@
...
-"TO interact with UC functions using serverless compute, install the package with "
-f"`pip install databricks-connect=={DATABRICKS_CONNECT_SUPPORTED_VERSION}`. "
-"Please note this requires python>=3.10."
-DATABRICKS_CONNECT_VERSION_NOT_SUPPORTED_ERROR_MESSAGE = "Serverless is not supported by the "
-"current databricks-connect version, install with "
-f"`pip install databricks-connect=={DATABRICKS_CONNECT_SUPPORTED_VERSION}` "
-"to use serverless compute in Databricks. Please note this requires python>=3.10."
+DATABRICKS_CONNECT_IMPORT_ERROR_MESSAGE = (
+    "Could not import databricks-connect python package. "
+    "TO interact with UC functions using serverless compute, install the package with "
```
```suggestion
    "To interact with UC functions using serverless compute, install the package with "
```

integrations/langchain/pyproject.toml
serena-ruan: It's used in the integrations test PR in the private repo

dev/databricks_functions_cleanup.py
```diff
@@ -0,0 +1,34 @@
...
+    failed_deletions = []
+    function_infos = client.functions.list(catalog_name=catalog, schema_name=schema)
+    for function_info in function_infos:
+        try:
+            client.functions.delete(function_info.full_name)
+        except Exception:
+            failed_deletions.append(function_info.full_name)
+
+    if failed_deletions:
+        print(f"Failed to delete the following functions: {failed_deletions}")  # noqa: T201
```
harupy:
```suggestion
        sys.stderr.write(f"Failed to delete the following functions: {failed_deletions}")  # noqa: T201
```

an error message should be written in stderr

harupy (approved):
LGTM!

dev/databricks_functions_cleanup.py
```diff
@@ -0,0 +1,34 @@
+import sys
+
+import click
```
harupy: Can I use `argparse` to avoid increasing dev dependencies. I'd consider using click if we had a few more CLI commands, or needed to write a more complex CLI.

dev/databricks_functions_cleanup.py
```diff
@@ -0,0 +1,34 @@
...
+)
+def cleanup_function(catalog: str, schema: str):
+    client = WorkspaceClient()
+    failed_deletions = []
+    function_infos = client.functions.list(catalog_name=catalog, schema_name=schema)
+    for function_info in function_infos:
+        try:
+            client.functions.delete(function_info.full_name)
+        except Exception:
+            failed_deletions.append(function_info.full_name)
```
harupy: Let's include the error info as well.

dev/databricks_functions_cleanup.py
harupy, line 1: where do we run this script?

serena-ruan: In databricks-eng/ai-oss-integration-tests-runner if you don't mind taking another look at that PR :)

dev/databricks_functions_cleanup.py
```diff
@@ -0,0 +1,34 @@
...
+    "--catalog",
+    required=True,
+    help=f"The catalog to clean up functions from",
+)
+@click.option(
+    "--schema",
+    required=True,
+    help=f"The schema to clean up functions from",
+)
+def cleanup_function(catalog: str, schema: str):
```
harupy:
```suggestion
def cleanup_functions(catalog: str, schema: str):
```

nit

dev/databricks_functions_cleanup.py
harupy: shouldn't we add this script in databricks-eng/ai-oss-integration-tests-runner?

serena-ruan: I think it doesn't matter? The real tests exist here so I added here, btw it depends on databricks-sdk so it's quite weird to put in that repo

harupy: If this script is in `databricks-eng/ai-oss-integration-tests-runner`, it's easy to test the script. We can run it in CI and make sure it works fine.

harupy: It's less-error prone when the code/script is defined and executed in the same place.

harupy: This is not a blocker, feel free to merge.

harupy (approved):
LGTM

dev/databricks_functions_cleanup.py
serena-ruan: I'll merge it :) If something is broken here, we still need to fix in this repo, and we can manually trigger the test using the PR's branch.

PR #22, opened 2024-10-02
Add callable support for create function

BenWilson2:
Adds support for defining a valid callable to a new client API for creating a function from a python func


README.md
```diff
@@ -60,6 +60,41 @@ $$
...
+    a: An integer that will have an additional integer added to it.
+    b: An integer that will be added to the 'a' argument.
+  
+  Returns:
+    int: The resulting sum of the arguments a + b
+  """
+  return a + b
+
+# Ensure that you fully explain what the function does and how it can be used.
+function_description = "A function that accepts to integers, sums them, and returns the sum result as an integer."
```
BenWilson2:
```suggestion
function_description = "A function that accepts two integers, sums them, and returns the sum result as an integer."
```

src/ucai/core/databricks.py
```diff
@@ -211,6 +212,34 @@ def create_function(
             return self.client.functions.create(function_info)
         raise ValueError("Either function_info or sql_function_body should be provided.")
 
+    @override
+    def create_python_function(self, *, func: Callable, func_comment: str, catalog: str, schema: str) -> "FunctionInfo":
+        """
+        Create a UC function from a Python function.
+
```
BenWilson2:
```suggestion
        Create a UC function from a Python function.
        
        Note that in order for comments to be automatically applied to your function's
        arguments, you must utilize Google Style Python Docstrings when writing your
        function. For more information and examples of valid Docstrings to use, you can
        visit https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
        
        Note that type hints must be applied to the arguments of your function and your
        function must have a return type designated. Failure to specify type hints will
        result in an exception.
        
```

src/ucai/core/databricks.py
```diff
@@ -211,6 +212,34 @@ def create_function(
             return self.client.functions.create(function_info)
         raise ValueError("Either function_info or sql_function_body should be provided.")
 
+    @override
+    def create_python_function(self, *, func: Callable, func_comment: str, catalog: str, schema: str) -> "FunctionInfo":
```
```suggestion
    def create_python_function(self, *, func: Callable[..., Any], func_comment: str, catalog: str, schema: str) -> "FunctionInfo":
```

to avoid an ambiguous type hint.

src/ucai/core/databricks.py
```diff
@@ -211,6 +212,34 @@ def create_function(
...
+        """
+        Create a UC function from a Python function.
+
+        Args:
+            func: The Python callable function to convert into a UDF.
```
harupy: I think we should add an example and clarify what the requirements are.

README.md
```diff
@@ -60,6 +60,41 @@ $$
...
+
+```python
+# Example function
+def add_numbers_together(a: int, b: int) -> int:
+  """
+  Utility for adding two integers together
+
+  Args:
+    a: An integer that will have an additional integer added to it.
+    b: An integer that will be added to the 'a' argument.
```
harupy: Does the current docstring parsing logic work if the arg section looks like this

```suggestion
    a: An integer that will have an additional integer added to it.
    The second description for a `:`.
    b (int): An integer that will be added to the 'a' argument.
```

README.md
```diff
@@ -60,6 +60,41 @@ $$
 function_info = client.create_function(sql_function_body=sql_body)
 ```
 
+#### Create a UC python function
+
+The client offers an API that simplifies the process of directly creating a python function to be registered as a UC function.
+
+To get the optimal results for GenAI tool definitions, it is highly recommended to:
+
+- Provide a full Python docstring in Google docstring format to your function definition. This API extracts the information about arguments so that GenAI tools
```
harupy:
```suggestion
- Provide a full Python docstring in [Google docstring format](URL) to your function definition. This API extracts the information about arguments so that GenAI tools
```

README.md
```diff
@@ -60,6 +60,41 @@ $$
...
+will have additional context about the tool's capabilities.
+- Ensure that the description of the function is as descriptive as possible to prevent a tool use call that infers capabilities of the tool purely based on the
+function's name.
+- Use a clear and descriptive naming convention for the function to aid in identification within Unity Catalog as well as for GenAI systems to understand
+additional context about a configured tool within an agentic application.
+- Use type hints. Failure to provide them will cause an exception to be raised.
```
harupy: What types are supported? Only primitives such as `int`?

src/ucai/core/databricks.py
BenWilson2: good idea! Will update to make it more clear

src/ucai/core/databricks.py
BenWilson2: I was thinking the same thing :) Even my suggestion above is too ambiguous about how to use this. Since this API is actually doing quite a bit of work, I'm thinking of adding a few examples to the docstring just to show options that are available (good, good with comments, bad with no type hints, bad with wrong format docstrings)

README.md
BenWilson2: This is a great callout! I'll add a test and make sure that the parser handles this properly! Thanks Haru!

README.md
BenWilson2: Ah, good idea! Since we're using mkdocs, markdown comment syntax should just work in the generated API docs :)

README.md
BenWilson2: I'll add the full list of what's permitted and what is guaranteed not to work - basically if it's in the core lib type declarations or in the typing module, we'll pick it up, but the mapping to Spark SQL types is restricted to the enumeration that I created. I'll add a section on that in the doc string and the README :)

README.md
harupy: I'm not sure if this conforms to the google docstring format tho.

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+    if 'return' not in type_hints:
+        raise ValueError(f"Return type for function '{func_name}' is not defined. Please provide a return type.")
+    return_type = type_hints.get('return')
+
+    docstring = inspect.getdoc(func) or ""
+    docstring_comments = parse_docstring(docstring)
+
+    sql_params = []
+    for param_name, _ in signature.parameters.items():
```
harupy: how should we handle `*args` or `**kwargs`?

tests/core/test_databricks_client.py
```diff
@@ -218,6 +224,124 @@ def function_with_table_output(func_name: str) -> FunctionInputOutput:
...
+
+    return PythonFunctionInputOutput(
+        func=func,
+        inputs=[{"s": datetime.date(2024, 8, 19)}, {"s": "2024-08-19"}],
+        output="2024-08-19",
+    )
+
+
+def python_function_with_map_input() -> PythonFunctionInputOutput:
+    def func(s: dict) -> str:
```
harupy: Is `dict[str, <non int type>]` supported? For example: `dict[str, int]`

src/ucai/core/utils/type_utils.py
```diff
@@ -86,3 +98,17 @@ def convert_timedelta_to_interval_str(time_val: datetime.timedelta) -> str:
...
     return f"INTERVAL '{days} {hours}:{minutes}:{seconds}.{microseconds}' DAY TO SECOND"
+
+
+def python_type_to_sql_type(py_type):
+    """Convert Python type to Unity Catalog SQL type."""
+    origin = get_origin(py_type)
+    if origin is dict:
+        return "MAP"
+    elif origin is list:
+        return "ARRAY"
```
harupy: do we need to set the element type?

src/ucai/core/client.py
```diff
@@ -43,6 +43,10 @@ def __init__(self, client: Any = None, **kwargs: Any) -> None:  # noqa: ARG002
     def create_function(self, *args: Any, **kwargs: Any) -> Any:
         """Create a function"""
 
+    @abstractmethod
+    def create_python_function(self, *args: Any, **kwargs: Any) -> Any:
+        """Create a Python function"""
```
harupy: IMO, `*args` and `**kwargs` make this abstract class useless. We should explicitly define parameters what `create_python_function` should take.

src/ucai/core/utils/callable_utils.py
BenWilson2: These are not allowed by UC. There are existing tests to validate, but I'll add explicit remarks in the API Docs and the README to clearly mention this incompatibility.

src/ucai/core/client.py
BenWilson2: forgot to update that! thanks :)

src/ucai/core/utils/type_utils.py
BenWilson2: yep and we need to recursively parse the container types. Updating with additional test coverage validation.

README.md
BenWilson2: Added proper parsing for type definition removal in docstrings and added tests for validation

README.md
```diff
@@ -60,6 +60,190 @@ $$
...
+    | `str`               | `STRING`           |
+    | `bool`              | `BOOLEAN`          |
+    | `Decimal`           | `DECIMAL`          |
+    | `datetime.date`     | `DATE`             |
+    | `datetime.datetime` | `TIMESTAMP`        |
+    | `list`              | `ARRAY`            |
+    | `tuple`             | `ARRAY`            |
+    | `dict`              | `MAP`              |
+
+- **Container Types**: If you use container types like `list`, `tuple`, or `dict`, you **must specify the inner types** explicitly. Unity Catalog does not support ambiguous types like `List` or `Dict`. For example:
```
harupy:
```suggestion
- **Container Types**: If you use collection types like `list`, `tuple`, or `dict`, you **must specify the inner types** explicitly. Unity Catalog does not support ambiguous types like `List` or `Dict`. For example:
```

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+    function_body_lines = source_lines[body_start_line:]
+
+    min_indent = min((len(line) - len(line.lstrip()) for line in function_body_lines if line.strip()), default=0)
+    
+    function_body = "\n".join([line[min_indent:] if len(line) >= min_indent else line for line in function_body_lines])
+
+    return function_body
+
+def parse_docstring(docstring: str) -> dict:
```
harupy:
```suggestion
def parse_docstring(docstring: str) -> dict[???, ???]:
```

what's the key and value type?

src/ucai/core/utils/callable_utils.py
harupy: We can reject them more explicitly using https://docs.python.org/3/library/inspect.html#inspect.Parameter.kind.

src/ucai/core/utils/type_utils.py
```diff
@@ -86,3 +99,70 @@ def convert_timedelta_to_interval_str(time_val: datetime.timedelta) -> str:
...
+            raise ValueError(f"Unsupported Python type: typing.List or typing.Tuple requires an element type.")
+        
+        # Get element type for List or Tuple (only one type should be allowed)
+        element_type, = get_args(py_type)
+        element_sql_type = python_type_to_sql_type(element_type)
+        return f"ARRAY<{element_sql_type}>"
+    
+    # Handle base types (non-generic)
+    if py_type in PYTHON_TO_SQL_TYPE_MAPPING:
+        return PYTHON_TO_SQL_TYPE_MAPPING[py_type]
```
harupy:
```suggestion
    if sql_type := PYTHON_TO_SQL_TYPE_MAPPING.get(py_type):
        return sql_type
```

src/ucai/core/utils/callable_utils.py
harupy: We can use a state machine like approach to improve the code. I haven't tested but it should illustrate the idea.

```python
from dataclasses import dataclass
from typing import Dict, List, Optional

@dataclass
class ParsedDocstring:
    args: Dict[str, str]
    returns: Optional[str] = None
    raises: Optional[str] = None

def parse_docstring(docstring: str) -> ParsedDocstring:
    """Parses the docstring to extract comments for parameters, return value, and exceptions raised."""
    parsed_docstring = ParsedDocstring(args={})
    state: Optional[str] = None  # Possible states: 'args', 'returns', 'raises'
    current_param: Optional[str] = None
    description_lines: List[str] = []

    for line in docstring.splitlines():
        line = line.strip()
        
        if line.startswith("Args:"):
            state = 'args'
            continue
        elif line.startswith("Returns:"):
            state = 'returns'
            continue
        elif line.startswith("Raises:"):
            state = 'raises'
            continue

        if state == 'args':
            if line and ":" in line:
                if current_param and description_lines:
                    parsed_docstring.args[current_param] = " ".join(description_lines).strip()
                
                param_part, comment_part = line.split(":", 1)
                current_param = param_part.split("(")[0].strip()
                description_lines = [comment_part.strip()]
            elif current_param:
                description_lines.append(line)

        elif state == 'returns':
            parsed_docstring.returns = line.strip()

        elif state == 'raises':
            if line:
                parsed_docstring.raises = line.strip()

    # Finalize last parameter if in 'args' state
    if state == 'args' and current_param and description_lines:
        parsed_docstring.args[current_param] = " ".join(description_lines).strip()

    return parsed_docstring

```

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    Args:
+        func: The Python callable function to convert into a UDF.
+        func_comment: A short description for the function.
+        catalog: The catalog name.
+        schema: The schema name.
+    Returns:
+        str: SQL statement for creating the UDF.
+    """
+    # unwrap staticmethod or classmethod for compatilibity with Python<=3.9
+    if isinstance(func, staticmethod) or isinstance(func, classmethod):
```
harupy:
```suggestion
    if isinstance(func, (staticmethod, classmethod)):
```

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+    if "Raises:" in docstring:
+        raises_part = docstring.split("Raises:")[1].strip()
+        parsed_comments['raises'] = raises_part
+
+    return parsed_comments
+
+
+
+def parse_dedented_source(func: Callable[..., Any]) -> tuple:
```
harupy:
```suggestion
def parse_dedented_source(func: Callable[..., Any]) -> tuple[???, list[str]]:
```

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    if isinstance(function_def.body[0], ast.Expr) and isinstance(function_def.body[0].value, ast.Constant):
+        body_start_line += len(function_def.body[0].value.s.splitlines())
+
+    function_body_lines = source_lines[body_start_line:]
+
+    min_indent = min((len(line) - len(line.lstrip()) for line in function_body_lines if line.strip()), default=0)
+    
+    function_body = "\n".join([line[min_indent:] if len(line) >= min_indent else line for line in function_body_lines])
+
+    return function_body
```
harupy, line 27: We just need to know the first non-docstring node, right? We can define a visitor by extending `ast.Visitor`.

src/ucai/core/databricks.py
```diff
@@ -211,6 +212,141 @@ def create_function(
...
+            ```python
+            def my_function(a: int, b: str) -> float:
+                return a + len(b)
+            ```
+            
+            - **Invalid function (missing type annotations)**:
+            ```python
+            def my_function(a, b):
+                return a + len(b)
+            ```
```
harupy, line 263: Does this render fine on docs?

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,142 @@
...
+import inspect
+from textwrap import dedent
+from typing import Any, Callable, get_type_hints
+
+from ucai.core.utils.type_utils import python_type_to_sql_type, validate_container_type
+
+FORBIDDEN_PARAMS = ['self', 'cls']
+
+def extract_function_body(func: Callable[..., Any]) -> str:
+    """Extracts the body of a function as a string without the signature or docstring, preserving indentation."""
```
harupy: Do we need to preserve indentation?

harupy: Can we just `textwrap.indent` the body when injecting it in `CREATE FUNCTION`?

src/ucai/core/client.py
```diff
@@ -43,6 +43,10 @@ def __init__(self, client: Any = None, **kwargs: Any) -> None:  # noqa: ARG002
     def create_function(self, *args: Any, **kwargs: Any) -> Any:
         """Create a function"""
 
+    @abstractmethod
+    def create_python_function(self, *, func: Callable[..., Any], func_comment: str, catalog: str, schema: str) -> Any:
```
harupy: Is `func_comment` required?

README.md
```diff
@@ -60,6 +60,190 @@ $$
...
+
+```python
+def process_data(ids: List[int], details: Dict[str, float]) -> List[str]:
```
harupy, line 92: Is a function with optional values supported?

```
def f(s: str = None)
```

tests/core/test_callable_utils.py
```diff
@@ -0,0 +1,880 @@
...
+
+        Args:
+            a: A plain list without inner types
+
+        Returns:
+            str: A string representation of the list
+        """
+        return str(a)
+
+    with pytest.raises(ValueError, match="Error in parameter 'a': type typing.List is not supported."):
```
harupy: Can we make this error message more actionable? What could fix the error?

src/ucai/core/utils/callable_utils.py
BenWilson2: Interesting idea! Adjusted the suggestion to fit within the design :)

src/ucai/core/utils/type_utils.py
BenWilson2: thanks!

src/ucai/core/utils/callable_utils.py
BenWilson2: ahhhh ironic that I'm rejecting these undefined collections and I forget to supply them. Thanks

src/ucai/core/utils/callable_utils.py
BenWilson2: Great point. Refactored the implementation to adjust for 2 vs 4 indentation and simplified this dedent/indent logic. Validated in DBR that the indentation works for function execution

src/ucai/core/client.py
BenWilson2: yeah, without it, the GenAI tools just sort of "guess" at what a tool is. In tutorials, we'll be making this VERY clear that these are needed to make high-quality agentic tools

README.md
BenWilson2: it is now :) Thanks for the call-out!!

tests/core/test_callable_utils.py
BenWilson2: yeah, that message is terrible. I refactored the error handling to provide guidance on collections since this was so vague and confusing (required looking at the stack trace, which is NOT good)

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,285 @@
...
+    Extracts the body of a function as a string without the signature or docstring,
+    dedents the code, and returns the indentation unit used in the function (e.g., 2 or 4 spaces).
+    """
+    source_lines, _ = inspect.getsourcelines(func)
+    dedented_source = dedent(''.join(source_lines))
+
+    parsed_source = ast.parse(dedented_source)
+    func_name = func.__name__
+
+    class FunctionBodyExtractor(ast.NodeVisitor):
```
harupy: we can pull this out from the function. Inner classs/functions make the code difficult to follow, confuse code linters , and are error-prone.

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,285 @@
...
+        str: SQL statement for creating the UDF.
+    """
+    func = unwrap_function(func)
+    func_name = func.__name__
+    signature = inspect.signature(func)
+    type_hints = get_type_hints(func)
+
+    sql_return_type = validate_return_type(func_name, type_hints)
+
+    docstring = inspect.getdoc(func) or ""
```
harupy: can we throw earlier if docstring is empty?

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,285 @@
+import ast
+import inspect
+from textwrap import dedent, indent
+from typing import Any, Callable, Union, get_type_hints
+
+from ucai.core.utils.type_utils import python_type_to_sql_type
+
+FORBIDDEN_PARAMS = ['self', 'cls']
+
+def parse_docstring(docstring: str) -> dict[str, str]:
```
harupy: I'd return a dataclass with `args`, `return`, and `raises` fields. dict should not be used if you know the field names beforehand. In this particular case, we know `args`, `return`, and `raises` are always present.

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,285 @@
...
+    sql_params_str = ', '.join(sql_params)
+    sql_body = f"""
+    CREATE OR REPLACE FUNCTION {catalog}.{schema}.{func_name}({sql_params_str})
+    RETURNS {sql_return_type}
+    LANGUAGE PYTHON
+    COMMENT '{func_comment}'
+    AS $$
+{indented_body}
+    $$;
+    """
```
harupy:
```suggestion
    sql_body = f"""
CREATE OR REPLACE FUNCTION {catalog}.{schema}.{func_name}({sql_params_str})
RETURNS {sql_return_type}
LANGUAGE PYTHON
COMMENT '{func_comment}'
AS $$
{indented_body}
$$;
"""
```

should'n we dedent?

src/ucai/core/client.py
harupy: Can we extract if from the docstring? or is there a reason that that needs to be isolated as `func_comment`?

README.md
```diff
@@ -60,6 +60,279 @@ $$
...
+- **Metadata**: Generated from the function’s docstrings and descriptions passed via the `func_comment`.
+- **Validation**: If the function or any of its components (e.g., types) are invalid, an exception will be raised.
+- **Strict Return Types**: If a function’s return type is not supported or improperly defined, an exception will be raised. For instance, return types such as `Union[int, str]` will cause an error.
+
+---
+
+##### Example Usage
+
+```python
+# Define the function to be registered
```
harupy, line 327:
```suggestion
from typing import Dict, List

# Define the function to be registered
```

to make it easy to run this code.

src/ucai/core/databricks.py
BenWilson2: We'll make rendering adjustments when we integrate docs rendering with the unitycatalog configuration setup for mkdocs on the main repo.

src/ucai/core/client.py
BenWilson2: After doing some bug bashing, this makes a ton of sense. It's really annoying to type out a description if I already have it in my docstring. 
I have:
- Raised a ValueError if the description isn't supplied in the docstring
- Remove the argument since it's redundant
- Added tests to confirm the behavior of not supplying / providing an empty string as the description raises a ValueError

src/ucai/core/utils/callable_utils.py
BenWilson2: good call

src/ucai/core/utils/callable_utils.py
BenWilson2: yep - updated all of the tests and I'm preserving the indentation of the body to align with the expected layout for UC python functions. Also confirmed that it doesn't matter about dedent / indent on the python function that is registered, provided that relative indentation is consistent and correct for the function body. The current implementation matches the tutorials here: https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-sql-function.html#create-python-functions

BenWilson2:
e2e test validation: https://ai-oss-ecosystem-integration-testing.cloud.databricks.com/?o=3272836215725701#notebook/3395377771238729/command/4350157102064216

harupy (approved):
LGTM!

src/ucai/core/databricks.py
```diff
@@ -215,6 +216,150 @@ def create_function(
...
+            FunctionInfo: Metadata about the created function, including its name and signature.
+        """
+
+        if not callable(func):
+            raise ValueError("The provided function is not callable.")
+
+        try:
+            sql_function_body = generate_sql_function_body(func, catalog, schema)
+        except Exception as e:
+            raise RuntimeError(f"Failed to generate SQL function body for {func.__name__}") from e
```
harupy: do we need this try-catch?

BenWilson2: haha I was just doing more e2e testing and was thinking the same thing... it's just a wrapper around a more informative exception :) I'm going to remove it

BenWilson2: @harupy Updated to remove the exception and also added better error messages for the return types (they were too vague for the collections, Union, and Any types). Error messages are much more clear about what is wrong and provide concrete guidance now.

src/ucai/core/client.py
```diff
@@ -43,6 +43,10 @@ def __init__(self, client: Any = None, **kwargs: Any) -> None:  # noqa: ARG002
...
+        """Create a Python function"""
```
harupy: can we add the parameter decscriptions?

src/ucai/core/client.py
harupy:
```suggestion
    def create_python_function(self, *, func: Callable[..., Any], func_comment: str, catalog: str, schema: str) -> None:
```

if this function returns nothing.

serena-ruan: Let's mark func_comment as Optional[str]=None?

README.md
```diff
@@ -60,6 +60,277 @@ $$
...
+    Args:
+        orders (Dict[str, List[int]]): A dictionary where keys are customer names and values are lists of order totals.
+
+    Returns:
+        str: A summary of the orders.
+    """
+    return ", ".join(f"{k}: {sum(v)}" for k, v in orders.items())
+
+# Provide a description for the function
+function_description = "Processes a dictionary of customer orders and returns a summary string."
```
serena-ruan:
```suggestion
```

src/ucai/core/client.py
serena-ruan:
```suggestion
    def create_python_function(self, *, func: Callable[..., Any], catalog: str, schema: str) -> Any:
```

src/ucai/core/databricks.py
```diff
@@ -215,6 +216,147 @@ def create_function(
...
+    @override
+    def create_python_function(self, *, func: Callable[..., Any], catalog: str, schema: str) -> "FunctionInfo":
+        """
+        Create a Unity Catalog (UC) function directly from a Python function.
+
+        This API allows you to convert a Python function into a Unity Catalog User-Defined Function (UDF). 
+        It automates the creation of UC functions while ensuring that the Python function meets certain 
+        criteria and adheres to best practices.
+
+        **Requirements:**
```
serena-ruan, line 231: This is awesome but seems too long for a function docstring? Can we add a TODO to replace this requirements part with a link to the doc once we have it?

src/ucai/core/databricks.py
```diff
@@ -215,6 +216,147 @@ def create_function(
...
+            | `int`               | `INTEGER`          |
+            | `float`             | `DOUBLE`           |
+            | `str`               | `STRING`           |
+            | `bool`              | `BOOLEAN`          |
+            | `Decimal`           | `DECIMAL`          |
+            | `datetime.date`     | `DATE`             |
+            | `datetime.datetime` | `TIMESTAMP`        |
+            | `list`              | `ARRAY`            |
+            | `tuple`             | `ARRAY`            |
+            | `dict`              | `MAP`              |
```
serena-ruan: Seems bytes: "BINARY" and datetime.timedelta: "INTERVAL DAY TO SECOND" are missing here; Same as README

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,361 @@
...
+    )
+
+def format_default_value(default: Any) -> str:
+    """Formats the default value for SQL."""
+    if default is None:
+        return 'NULL'
+    elif isinstance(default, str):
+        return f"'{default}'"
+    else:
+        return str(default)
```
serena-ruan, line 229: Should we forbidden array and map default values?

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,361 @@
...
+    except ValueError as e:
+        error_message = generate_type_hint_error_message(param_name, param_hint, e)
+        raise ValueError(error_message) from e
+
+    param_comment = docstring_info.params.get(param_name, f"Parameter {param_name}").replace("'", '"')
+
+    if param.default is inspect.Parameter.empty:
+        return f"{param_name} {sql_type} COMMENT '{param_comment}'"
+    else:
+        default_value = format_default_value(param.default)
```
serena-ruan, line 320: Should we validate the param.default value is of correct type? 
For example, below code actually works:
```
def test(a: List[str]=1):
    return a
test()
```

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,361 @@
...
+    func_name: str,
+    sql_params: list[str],
+    sql_return_type: str,
+    func_comment: str,
+    indented_body: str
+) -> str:
+    """Assembles the final SQL function body."""
+    sql_params_str = ', '.join(sql_params)
+    sql_body = f"""
+CREATE OR REPLACE FUNCTION {catalog}.{schema}.{func_name}({sql_params_str})
```
serena-ruan: We might want to add a param `override=False` to control whether to replace the function if it already exists. By default adding `OR REPLACE` sounds dangerous.

src/ucai/test_utils/function_utils.py
```diff
@@ -75,3 +82,24 @@ def create_function_and_cleanup(
...
+            full_function_name=func_name,
+            comment=func_info.comment,
+            tool_name=get_tool_name(func_name),
+        )
+    finally:
+        try:
+            client.client.functions.delete(func_name)
+        except Exception as e:
+            _logger.warning(f"Fail to delete function: {e}"
+)
```
serena-ruan:
```suggestion
)

```

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,361 @@
...
+        raise ValueError("Docstring is empty. Please provide a docstring with a function description.")
+
+    description_lines = []
+    parsed_params = {}
+    returns = None
+    raises = None
+    current_param = None
+    param_description_lines = []
+
+    class State:
```
serena-ruan: Can we move this class outside of this function?

src/ucai/core/utils/callable_utils.py
```diff
@@ -0,0 +1,361 @@
...
+FORBIDDEN_PARAMS = ['self', 'cls']
+
+
+@dataclass
+class DocstringInfo:
+    """Dataclass to store parsed docstring information."""
+    description: str
+    params: Optional[dict[str, str]]
+    returns: Optional[str]
+    raises: Optional[str]
```
serena-ruan: How is `raises` used? I didn't find it 🤔

tests/core/test_callable_utils.py
```diff
@@ -0,0 +1,1155 @@
...
+        return str(a)
+
+    with pytest.raises(ValueError, match=re.escape("Error in return type for function 'func_returning_union': typing.Union[int, str]. Union types are not supported in return types.")):
+        generate_sql_function_body(func_returning_union, "test_catalog", "test_schema")
+
+# ---------------------------
+# Tests for Input Types
+# ---------------------------
+
+def test_function_with_list_input():
```
serena-ruan, line 493: As we're only testing different input types, we can parametrize them? Not a blocker, we can refactor in a follow-up PR, this PR is too much :)

src/ucai/core/databricks.py
BenWilson2: Ah thanks for the catch :)

src/ucai/core/databricks.py
BenWilson2: yep!

src/ucai/core/client.py
BenWilson2: forgot to remove that one :)

BenWilson2: deleted

BenWilson2: Updated to return FunctionInfo

src/ucai/core/utils/callable_utils.py
BenWilson2: Yeah good call. Defaults in collections are always odd. Added validation and tests to confirm

src/ucai/core/utils/callable_utils.py
BenWilson2: great idea! Added :)

tests/core/test_callable_utils.py
BenWilson2: Yeah, I'm voting for a followup for optimizing tests

src/ucai/core/utils/callable_utils.py
BenWilson2: Removed this as the code that used it is now dead

PR #26, opened 2024-10-08
Add env variables for controlling execution args

serena-ruan:
The old way for controlling execute_statement arguments is too complex and hard for LLM to understand, using environment variable to control is much simpler and easier.


serena-ruan:
image

src/ucai/core/envs/databricks_env_vars.py
```diff
@@ -0,0 +1,27 @@
...
+
+EXECUTE_FUNCTION_WAIT_TIMEOUT = _EnvironmentVariable("EXECUTE_FUNCTION_WAIT_TIMEOUT", "30s")
+EXECUTE_FUNCTION_ROW_LIMIT = _EnvironmentVariable("EXECUTE_FUNCTION_ROW_LIMIT", "100")
+EXECUTE_FUNCTION_BYTE_LIMIT = _EnvironmentVariable("EXECUTE_FUNCTION_BYTE_LIMIT", "4096")
+UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT = _EnvironmentVariable(
+    "UNITYCATALOG_AI_CLIENT_EXECUTION_TIMEOUT", "120"
+)
+UC_AI_CLIENT_EXECUTION_RESULT_ROW_LIMIT = _EnvironmentVariable(
+    "UC_AI_CLIENT_EXECUTION_RESULT_ROW_LIMIT", "100"
+)
```
harupy: can we add what these environment variables are for?

src/ucai/core/envs/databricks_env_vars.py
```diff
@@ -0,0 +1,27 @@
...
+        self.default_value = str(default_value)
+
+    def get(self):
+        return os.getenv(self.name, self.default_value)
+
+    def set(self, value):
+        os.environ[self.name] = str(value)
+
+    def remove(self):
+        os.environ.pop(self.name, None)
```
harupy: can you add type hints?

src/ucai/core/envs/databricks_env_vars.py
```diff
@@ -0,0 +1,55 @@
...
+    "Maximum number of rows of the function execution result using Databricks client with warehouse_id; "
+    "it also sets the `truncated` field in the response to indicate whether the result was trimmed due to the limit or not.",
+)
+UCAI_DATABRICKS_WAREHOUSE_EXECUTE_FUNCTION_BYTE_LIMIT = _EnvironmentVariable(
+    "UCAI_DATABRICKS_WAREHOUSE_EXECUTE_FUNCTION_BYTE_LIMIT",
+    "4096",
+    "Maximum byte size of the function execution result using Databricks client with warehouse_id; "
+    "If the result was truncated due to the byte limit, then `truncated` in the response is set to `true`.",
+)
+UCAI_DATABRICKS_WAREHOUSE_RETRY_TIMEOUT = _EnvironmentVariable(
```
BenWilson2, line 43: Instead of handling two temporal settings for these timeouts, would it be a little less confusing to have a retry count that iterates on `UCAI_DATABRICKS_WAREHOUSE_EXECUTE_FUNCTION_WAIT_TIMEOUT`? 

I'm curious what the behavior would be if I set that wait timeout value to say 45s, the RETRY timeout to 60... would I effectively abort on the first retry if the total execution time of my function is 20s, even if it was going to return, or do I only abort a retry if the total elapsed time exceeds this value? 
Would it be more intuitive to just say "I want to retry at max 3 times where if I'm waiting for more than 20s on each one, I want to retry with some delay between the retries"?

serena-ruan: If the statement execution completes within `WAIT_TIMEOUT` then the response state should be SUCCEEDED, and it won't enter the retry loop; We only retry if the statement is still executing after `WAIT_TIMEOUT`, and try to fetch the response until `RETRY_TIMEOUT` is reached.
Sometimes people could set WAIT_TIMEOUT to a large number as it's the total execution time, but we might not want to wait the same time for retrieving response? Exponential retry time seems make more sense 🤔 
If this is confusing, I can add documentation in the README to explain them!

src/ucai/core/envs/databricks_env_vars.py
```diff
@@ -0,0 +1,55 @@
+import os
+from typing import Optional
+
+
+class _EnvironmentVariable:
+    def __init__(self, name: str, default_value: str, description: Optional[str] = None):
```
harupy:
```suggestion
    def __init__(self, name: str, default_value: str, description: str):
```

I think description should not be optional.

README.md
```diff
@@ -371,6 +371,18 @@ result = client.execute_function(full_func_name, {"s": "some_string"})
...
+  - `UCAI_DATABRICKS_WAREHOUSE_EXECUTE_FUNCTION_BYTE_LIMIT`: Maximum byte size of the function execution result; if the result was truncated due to the byte limit, then `truncated` in the response is set to `true`. Default value is 4096.
+  - `UCAI_DATABRICKS_WAREHOUSE_RETRY_TIMEOUT`: Client side retry timeout for the function execution; if the function execution does not complete within UCAI_DATABRICKS_WAREHOUSE_EXECUTE_FUNCTION_WAIT_TIMEOUT, client side will retry to fetch the result until this timeout is reached. Default value is 120. e.g. The function execution is still running after 30s (default value of wait_timeout), then we retry to fetch the response with exponential wait time 2s, 4s, ... until `UCAI_DATABRICKS_WAREHOUSE_RETRY_TIMEOUT` seconds reach.
+- If the client uses serverless compute for function execution:
+  - `UCAI_DATABRICKS_SERVERLESS_EXECUTION_RESULT_ROW_LIMIT`: Maximum number of rows of the function execution result. Default value is 100.
```
harupy: I think this will be outdated easily. Not a blocker, but can we automate the process to update this list?

README.md
```diff
@@ -371,6 +371,18 @@ result = client.execute_function(full_func_name, {"s": "some_string"})
 assert result.value == "some_string"
 ```
 
+##### Function execution arguments configuration
```
harupy, line 374: This is what it looks like:

image

It's a bit challenging to tell what environment variables are available. I'd use a table.

src/ucai/core/client.py
```diff
@@ -43,13 +40,13 @@ def __init__(self, client: Any = None, **kwargs: Any) -> None:  # noqa: ARG002
...
     @abstractmethod
-    def create_function(self, *args: Any, **kwargs: Any) -> "FunctionInfo":
+    def create_function(self, *args: Any, **kwargs: Any) -> Any:
         """Create a function"""
 
     @abstractmethod
     def create_python_function(
         self, *, func: Callable[..., Any], catalog: str, schema: str, replace: bool = False
-    ) -> "FunctionInfo":
+    ) -> Any:
```
harupy, line 49: why Any?

serena-ruan: OSS client will have a different FunctionInfo type, and we shouldn't introduce databricks as a dependency in base client

README.md
serena-ruan: Make sense, we should add a script to update it

README.md
serena-ruan: Updated! 
image

src/ucai/core/client.py
harupy: I see. This makes it hard to use client. For example:

```python
def some_func(client: BaseClient):
    func = create_python_function(...)
    
    # As func is Any, we need to check what it is using isinstance or getattr
    if isinstance(func, Foo):
        ...
    elif ...:
        ...
```

harupy: > OSS client will have a different FunctionInfo type

Can we define a common daataclass for that?

harupy: A change like this should be separated to avoid unnecessary discussion or blockers.

serena-ruan: As this is an abstract method, the `real` client should override this method and update the typehint

serena-ruan: FunctionInfo is super complex and nested, so I would prefer that we don't wrap a dataclass on it

harupy: I see, so the client consumer like `some_func` above needs to inspect the object?

serena-ruan: Well it depends on which client it uses, if using databricks client then func must be FunctionInfo type, if using OSS then func must be unitycatalog.types.function_info.FunctionInfo type. They should share the same structure, but this sdk shouldn't define them.
If `some_func` depends on general BaseFunctionClient, then it has to figure out what instance type create_function returns.

harupy: What about the integrations in this repo? Not sure if they touch `create_python_function`, but if they do, do they need to figure out what `create_python_function` returns as `some_func` does?

harupy: Not a blocker, but returning `Any` is like "What is this abstract class for?" and requires its consumer to guess what the return value is, which is error-prone.

serena-ruan: Yes, the clients in this repo (Databricks & OSS) should override create_python_function and update the return type.  Databricks already did by updating the return type as FunctionInfo

harupy (approved):
LGTM

PR #27, opened 2024-10-08
Add langgraph test

serena-ruan:
Langgraph supports the same tool as Langchain, so we don't need extra integration with it.
Adding a test for langgraph.
image

### DIFF
--- a/integrations/langchain/tests/test_toolkit.py
+++ b/integrations/langchain/tests/test_toolkit.py
@@ -8,6 +8,11 @@ from databricks.sdk.service.catalog import (
     FunctionParameterInfo,
     FunctionParameterInfos,
 )
+from langchain_core.messages import BaseMessage
+from langchain_core.outputs import ChatGeneration, ChatResult
+from langchain_core.runnables import RunnableGenerator
+from langchain_databricks.chat_models import ChatDatabricks, ChatGeneration
+from langgraph.prebuilt import create_react_agent
 from ucai.core.client import (
     FunctionExecutionResult,
 )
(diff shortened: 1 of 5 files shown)


integrations/langchain/tests/test_toolkit.py
```diff
@@ -159,3 +164,34 @@ def test_uc_function_to_langchain_tool():
...
+def test_langgraph_agents(monkeypatch, use_serverless):
+    monkeypatch.setenv(USE_SERVERLESS, str(use_serverless))
+    client = get_client()
+    with create_function_and_cleanup(client, schema=SCHEMA) as func_obj:
+        toolkit = UCFunctionToolkit(function_names=[func_obj.full_function_name], client=client)
+        system_message = "You are a helpful assistant. Make sure to use tool for information."
+        llm = ChatDatabricks(endpoint="databricks-meta-llama-3-1-70b-instruct")
+        agent = create_react_agent(llm, toolkit.tools, state_modifier=system_message)
+        chain = agent | RunnableGenerator(wrap_output)
+        # we need this mock as the serving endpoint is not stable
```
BenWilson2, line 180: Are all of the FMAPI endpoints unstable on our workspace or just large models like the 70b variant?

BenWilson2 (approved):
Looks straightforward. Thanks for adding coverage for LangGraph!

integrations/langchain/tests/test_toolkit.py
serena-ruan: I tried 70b and 405b, they're not generating correct results. But OpenAI works fine, so I just mock the response lol

PR #29, opened 2024-10-09
Update LlamaIndex for tool schema validation

BenWilson2:
LlamaIndex requires the pydantic model definition to be passed for parameter signature validation, rather than an instance. This PR updates the logic to pass the correct type.

### DIFF
--- a/integrations/llama_index/src/ucai_llamaindex/toolkit.py
+++ b/integrations/llama_index/src/ucai_llamaindex/toolkit.py
@@ -129,18 +129,23 @@ class UCFunctionToolkit(BaseModel):
             raise ValueError("Either function_name or function_info should be provided.")
 
         fn_schema = generate_function_input_params_schema(function_info)
+        # LlamaIndex requires the Pydantic model definition rather than an instance for validation
+        pydantic_model = fn_schema.pydantic_model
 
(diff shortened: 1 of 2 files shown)


BenWilson2:
An alternative implementation that addresses the same breaking change issue in LlamaIndex is here: https://github.com/serena-ruan/unitycatalog-ai/pull/30

integrations/llama_index/src/ucai_llamaindex/toolkit.py
```diff
@@ -129,28 +114,39 @@ def uc_function_to_llama_tool(
             raise ValueError("Either function_name or function_info should be provided.")
 
         fn_schema = generate_function_input_params_schema(function_info)
+        pydantic_model = fn_schema.pydantic_model
+
+        # Enforce strict validation by setting 'extra' to 'forbid' for both outer and inner models
+        WrappedModel = create_model(
```
BenWilson2, line 120: I'm not the biggest fan of this extreme restriction. I feel like this is going to be an annoying maintenance burden if LlamaIndex modifies the internal data structures again.

serena-ruan:
I agree https://github.com/serena-ruan/unitycatalog-ai/pull/30 is a better solution :)

PR #30, opened 2024-10-09
simplify the conversion safely

BenWilson2:
With the LangChain change in the Pydantic tool calling functionality to put tool call arguments within a `properties` argument as a dictionary, this is an alternate approach to handling the extraction of the arguments to the top-level for args processing. 
Prior to the recent changes and in testing of this fix, the Agent OCCASIONALLY generated a tool call as it formerly had within the same REPL instance. 

before:

tool call => `{"x": "print('hello world')"}`

recent update:
(this seems to be the default behavior and what each tool call is initiated with on the first attempt to call a tool) 

tool call => `{"properties": AttributedDict([("x", "print('hello world')")])}`

The approach here is to simply flatten the pydantic field extraction by popping `properties` if there are no key collisions with the remainder of the uc function definition payload for the function schema. The alternative (which has also been tested as working properly) is here: https://github.com/serena-ruan/unitycatalog-ai/pull/29 

For backwards compatibility and for flexibility in the direction of where LlamaIndex seems to be going for tool calling, I feel that this approach seems a bit better.

### DIFF
--- a/integrations/llama_index/src/ucai_llamaindex/toolkit.py
+++ b/integrations/llama_index/src/ucai_llamaindex/toolkit.py
@@ -90,7 +90,15 @@ class UCFunctionToolkit(BaseModel):
             uc_function_to_tool_func=self.uc_function_to_llama_tool,
             return_direct=self.return_direct,
         )
-        return self
+
+        # Since the 'properties' key is a reserved arg in LlamaIndex, disallow creating a tool with a
+        # function that has a 'properties' key in its input schema for LlamaIndex tool usage.
+        for tool_name, tool in self.tools_dict.items():
+            if "properties" in tool.metadata.fn_schema.model_fields:
+                raise ValueError(
+                    f"Function '{tool_name}' has a 'properties' key in its input schema. "
+                    "Cannot create a tool with this function due to LlamaIndex reserving this argument name."
+                )
 
     @staticmethod
     def uc_function_to_llama_tool(
@@ -199,9 +207,7 @@ def extract_properties(data: dict) -> dict:
         raise TypeError("'properties' must be a dictionary.")
 
     top_level_keys = set(data.keys()) - {"properties"}
-
     properties_keys = set(properties.keys())
-
     overlapping_keys = top_level_keys & properties_keys
 
     if overlapping_keys:
--- a/integrations/llama_index/src/ucai_llamaindex/toolkit.py
+++ b/integrations/llama_index/src/ucai_llamaindex/toolkit.py
@@ -140,6 +140,7 @@ class UCFunctionToolkit(BaseModel):
             Returns:
                 str: The JSON result of the function execution.
             """
+            kwargs = extract_properties(kwargs)
             args_json = json.loads(json.dumps(kwargs, default=str))
             result = client.execute_function(
                 function_name=function_name,
@@ -150,7 +151,7 @@ class UCFunctionToolkit(BaseModel):
         metadata = ToolMetadata(
             name=get_tool_name(function_name),
             description=function_info.comment or "",
-            fn_schema=fn_schema,
+            fn_schema=fn_schema.pydantic_model,
             return_direct=return_direct,
         )
 
@@ -169,3 +170,48 @@ class UCFunctionToolkit(BaseModel):
             List[FunctionTool]: A list of tools available in the toolkit.
         """
         return list(self.tools_dict.values())
+
+
+def extract_properties(data: dict) -> dict:
+    """
+    Extracts the 'properties' dictionary from the input dictionary,
+    merges its key-value pairs into the top-level dictionary, and returns a new dictionary.
+
+    Args:
+        data (dict): The original dictionary possibly containing a 'properties' key.
+
+    Returns:
+        dict: A new dictionary with 'properties' merged into the top-level.
+
+    Raises:
+        TypeError: If 'properties' exists but is not a dictionary.
+        KeyError: If there are key collisions between 'properties' and the top-level keys.
+    """
+    if not isinstance(data, dict):
+        raise TypeError("Input must be a dictionary.")
+
+    if "properties" not in data:
+        return data.copy()
+
+    properties = data["properties"]
+
+    if not isinstance(properties, dict):
+        raise TypeError("'properties' must be a dictionary.")
+
+    top_level_keys = set(data.keys()) - {"properties"}
+
+    properties_keys = set(properties.keys())
+
+    overlapping_keys = top_level_keys & properties_keys
+
+    if overlapping_keys:
+        conflict_keys = ", ".join(overlapping_keys)
+        raise KeyError(
+            f"Key collision detected for keys: {conflict_keys}. Cannot merge 'properties'."
+        )
+
+    merged_data = {k: v for k, v in data.items() if k != "properties"}
+
+    merged_data.update(properties)
+
+    return merged_data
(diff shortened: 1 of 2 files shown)


integrations/llama_index/src/ucai_llamaindex/toolkit.py
```diff
@@ -169,3 +170,48 @@ def tools(self) -> List[FunctionTool]:
             List[FunctionTool]: A list of tools available in the toolkit.
         """
         return list(self.tools_dict.values())
+
+
+def extract_properties(data: dict) -> dict:
+    """
+    Extracts the 'properties' dictionary from the input dictionary,
```
serena-ruan, line 186: What if the function has a key named `properties`?

serena-ruan: I think we need an extra layer of validation here, using client to fetch the function info and see if the function input params contains `properties`, and check the data type to see if it matches.

BenWilson2: Instead of performing a collision check by making another call, I switched to blocking the registration of a uc python function that contains the "properties" field on a registered function. 
I was a bit fed up with the tests, so I've added mocks so this can be locally validated and added unit test equivalents for the e2e tests that require creating functions in UC.

integrations/llama_index/src/ucai_llamaindex/toolkit.py
```diff
@@ -90,7 +90,15 @@ def validate_toolkit(self) -> "UCFunctionToolkit":
             uc_function_to_tool_func=self.uc_function_to_llama_tool,
             return_direct=self.return_direct,
         )
-        return self
+
+        # Since the 'properties' key is a reserved arg in LlamaIndex, disallow creating a tool with a
+        # function that has a 'properties' key in its input schema for LlamaIndex tool usage.
+        for tool_name, tool in self.tools_dict.items():
```
BenWilson2, line 96: Blocking logic

integrations/llama_index/src/ucai_llamaindex/toolkit.py
```diff
@@ -169,3 +178,46 @@ def tools(self) -> List[FunctionTool]:
...
+        TypeError: If 'properties' exists but is not a dictionary.
+        KeyError: If there are key collisions between 'properties' and the top-level keys.
+    """
+    if not isinstance(data, dict):
+        raise TypeError("Input must be a dictionary.")
+
+    if "properties" not in data:
+        return data.copy()
+
+    properties = data["properties"]
```
harupy: any chance that `properties` is None?

integrations/llama_index/src/ucai_llamaindex/toolkit.py
```diff
@@ -169,3 +178,46 @@ def tools(self) -> List[FunctionTool]:
...
+
+    Raises:
+        TypeError: If 'properties' exists but is not a dictionary.
+        KeyError: If there are key collisions between 'properties' and the top-level keys.
+    """
+    if not isinstance(data, dict):
+        raise TypeError("Input must be a dictionary.")
+
+    if "properties" not in data:
+        return data.copy()
```
harupy:
```suggestion
        return data
```

why copy?

integrations/llama_index/src/ucai_llamaindex/toolkit.py
```diff
@@ -169,3 +178,46 @@ def tools(self) -> List[FunctionTool]:
...
+    properties = data["properties"]
+
+    if not isinstance(properties, dict):
+        raise TypeError("'properties' must be a dictionary.")
+
+    top_level_keys = set(data.keys()) - {"properties"}
+    properties_keys = set(properties.keys())
+    overlapping_keys = top_level_keys & properties_keys
+
+    if overlapping_keys:
```
harupy:
```suggestion
    if overlapping_keys := top_level_keys & properties_keys:
```

integrations/llama_index/tests/test_toolkit.py
```diff
@@ -176,3 +269,234 @@ def test_toolkit_with_invalid_function_input(client):
...
+
+    with (
+        mock.patch(
+            "ucai.core.utils.function_processing_utils.generate_function_input_params_schema",
+            return_value=mock_fn_schema,
+        ),
+        mock.patch(
+            "ucai_llamaindex.toolkit.validate_or_set_default_client",
+            return_value=mock_client,
+        ),
```
harupy: can we validate these mock are used using `assert_called`? Imagine we make updates, and `generate_function_input_params_schema` is no longer called. That's a useless mock. `assert_called` can prevent it.

integrations/llama_index/src/ucai_llamaindex/toolkit.py
```diff
@@ -169,3 +178,46 @@ def tools(self) -> List[FunctionTool]:
...
+
+    Returns:
+        dict: A new dictionary with 'properties' merged into the top-level.
+
+    Raises:
+        TypeError: If 'properties' exists but is not a dictionary.
+        KeyError: If there are key collisions between 'properties' and the top-level keys.
+    """
+    if not isinstance(data, dict):
+        raise TypeError("Input must be a dictionary.")
```
harupy:
```suggestion
        raise TypeError("Input must be a dictionary. {type(data).__name__}")
```

The actual data type should be included. It helps debugging.

integrations/llama_index/src/ucai_llamaindex/toolkit.py
```diff
@@ -169,3 +178,46 @@ def tools(self) -> List[FunctionTool]:
             List[FunctionTool]: A list of tools available in the toolkit.
         """
         return list(self.tools_dict.values())
+
+
+def extract_properties(data: dict) -> dict:
```
harupy:
```suggestion
def extract_properties(data: dict[???, ???]) -> dict[???, ???]:
```

Can we add key and value types? When we use collection types, we should set their child types (element type, key type, value type) to eliminate ambiguity. `dict[Any, Any]` (explicit) is much better than `dict` (implicit).

integrations/llama_index/tests/test_toolkit.py
```diff
@@ -176,3 +269,234 @@ def test_toolkit_with_invalid_function_input(client):
...
+    }
+
+    assert extract_properties(data) == expected
+
+
+def test_extract_properties_non_dict_input():
+    data = ["not", "a", "dict"]
+
+    with pytest.raises(TypeError, match="Input must be a dictionary."):
+        extract_properties(data)
```
harupy:
```suggestion
def test_extract_properties_non_dict_input():
    with pytest.raises(TypeError, match="Input must be a dictionary."):
        extract_properties(data=0)
```

integrations/llama_index/tests/test_toolkit.py
```diff
@@ -176,3 +269,234 @@ def test_toolkit_with_invalid_function_input(client):
...
+        "properties": {"location": "abc", "metadata": "conflict", "name": "conflict_name"},
+        "metadata": "something",
+        "name": "something else",
+    }
+
+    with pytest.raises(
+        KeyError,
+        match="Key collision detected for keys",
+    ):
+        extract_properties(data)
```
harupy: can we parameterize these two tests?

serena-ruan (approved):
LGTM! Good to merge after addressing nits from Haru :)

integrations/llama_index/src/ucai_llamaindex/toolkit.py
BenWilson2: oooooh good catch. We definitely need to validate that. Thanks Haru!

integrations/llama_index/src/ucai_llamaindex/toolkit.py
BenWilson2: Artifact of debugging. Removing!

integrations/llama_index/tests/test_toolkit.py
BenWilson2: yep!

BenWilson2: great call - we preempt the patch for generating the schema (no need to patch as it is unreachable). Remove the patch (and added call assertions for the client patch)

PR #31, opened 2024-10-10
add warnings for docstrings formatting

BenWilson2:
Adds warnings to improperly formatted and constructed docstrings so that users can be made aware of creating a tool that is not optimal for GenAI tool usage.

### DIFF
--- a/src/ucai/core/utils/callable_utils.py
+++ b/src/ucai/core/utils/callable_utils.py
@@ -1,8 +1,9 @@
 import ast
 import inspect
+import warnings
 from dataclasses import dataclass
 from textwrap import dedent, indent
-from typing import Any, Callable, Optional, Union, get_args, get_origin, get_type_hints
+from typing import Any, Callable, Optional, Set, Union, get_args, get_origin, get_type_hints
 
 from ucai.core.utils.type_utils import python_type_to_sql_type
 
@@ -374,6 +375,9 @@ def generate_sql_function_body(
         raise ValueError(f"Function '{func_name}' must have a docstring with a description.")
     docstring_info = parse_docstring(docstring)
 
+    params_in_signature = set(signature.parameters.keys()) - set(FORBIDDEN_PARAMS)
+    check_docstring_signature_consistency(docstring_info.params, params_in_signature, func_name)
+
     sql_params = []
     for param_name, param in signature.parameters.items():
         sql_param = process_parameter(param_name, param, type_hints, docstring_info)
@@ -428,3 +432,49 @@ def validate_return_type(func_name: str, type_hints: dict[str, Any]) -> str:
 
         raise ValueError(base_msg) from e
     return sql_return_type
+
+
+def check_docstring_signature_consistency(
+    doc_params: Optional[dict[str, str]], signature_params: Set[str], func_name: str
+) -> None:
+    """
+    Checks for inconsistencies between docstring parameters and function signature parameters.
+    Issues warnings if there are mismatches.
+
+    Args:
+        doc_params (Optional[dict[str, str]]): Parameters documented in the docstring.
+        signature_params (Set[str]): Parameters present in the function signature.
+        func_name (str): The name of the function being checked.
+
+    Returns:
+        None
+    """
+    params_in_doc = set(doc_params.keys()) if doc_params else set()
+    params_in_signature = signature_params
+
+    extra_in_doc = params_in_doc - params_in_signature
+    extra_in_signature = params_in_signature - params_in_doc
+
+    if extra_in_doc:
+        warnings.warn(
+            f"In function '{func_name}': The following parameters are documented in the docstring but not present in the function signature: {', '.join(sorted(extra_in_doc))}",
+            UserWarning,
+        )
+
+    if extra_in_signature:
+        warnings.warn(
+            f"In function '{func_name}': The following parameters are present in the function signature but not documented in the docstring: {', '.join(sorted(extra_in_signature))}",
+            UserWarning,
+        )
+
+    if doc_params and not signature_params:
+        warnings.warn(
+            f"In function '{func_name}': Docstring defines parameters, but the function has no parameters in its signature.",
+            UserWarning,
+        )
+
+    if not doc_params and signature_params:
+        warnings.warn(
+            f"In function '{func_name}': Function has parameters in its signature, but the docstring does not document any parameters.",
+            UserWarning,
+        )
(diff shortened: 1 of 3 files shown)


src/ucai/core/utils/callable_utils.py
```diff
@@ -428,3 +432,53 @@ def validate_return_type(func_name: str, type_hints: dict[str, Any]) -> str:
 
         raise ValueError(base_msg) from e
     return sql_return_type
+
+
+def check_docstring_signature_consistency(
+    doc_params: Optional[dict[str, str]], signature_params: Set[str], func_name: str
```
serena-ruan, line 438: Should signature_params be Optional[Set[str]]?

tests/core/test_callable_utils.py
```diff
@@ -155,7 +160,7 @@ def my_multiline_arg_function(a: int, b: str) -> str:
     assert sql_body.strip() == expected_sql.strip()
 
 
-def test_function_with_extra_docstring_params_ignored():
+def test_function_with_extra_docstring_params_ignored(recwarn):
```
serena-ruan: Woo this is cool!

serena-ruan (approved):
LGTM!

src/ucai/core/utils/callable_utils.py
```diff
@@ -428,3 +432,53 @@ def validate_return_type(func_name: str, type_hints: dict[str, Any]) -> str:
...
+    extra_in_signature = params_in_signature - params_in_doc
+
+    if extra_in_doc:
+        warnings.warn(
+            f"In function '{func_name}': The following parameters are documented in the docstring but not present in the function signature: {', '.join(sorted(extra_in_doc))}",
+            UserWarning,
+            stacklevel=2,
+        )
+
+    if extra_in_signature:
```
harupy:
```suggestion
    if extra_in_doc := params_in_doc - params_in_signature
        warnings.warn(
            f"In function '{func_name}': The following parameters are documented in the docstring but not present in the function signature: {', '.join(sorted(extra_in_doc))}",
            UserWarning,
            stacklevel=2,
        )

    if extra_in_signature := params_in_signature - params_in_doc
```

to minimize the lifetime of variables.

tests/core/test_callable_utils.py
```diff
@@ -185,6 +190,12 @@ def func_with_extra_param_in_docstring(a: int) -> str:
     """
 
     assert sql_body.strip() == expected_sql.strip()
+    assert len(recwarn) == 1
+    warn = recwarn.pop(UserWarning)
+    assert (
+        "The following parameters are documented in the docstring but not present in the function signature: b"
+        in str(warn.message)
+    )
```
harupy: Can we `pytest.warns` instead?

src/ucai/core/utils/callable_utils.py
BenWilson2: It will always be a `Set`, albeit an empty set in the callee location with the definition of:

`params_in_signature = set(signature.parameters.keys()) - set(FORBIDDEN_PARAMS)`

Which is equivalent to:

```python
a = set() - set([1,2,3])
print(a)
```

```bash
set()
```

src/ucai/core/utils/callable_utils.py
```diff
@@ -428,3 +432,53 @@ def validate_return_type(func_name: str, type_hints: dict[str, Any]) -> str:
...
+
+    Args:
+        doc_params (Optional[dict[str, str]]): Parameters documented in the docstring.
+        signature_params (Set[str]): Parameters present in the function signature.
+        func_name (str): The name of the function being checked.
+
+    Returns:
+        None
+    """
+    params_in_doc = set(doc_params.keys()) if doc_params else set()
```
harupy:
```suggestion
    params_in_doc = set(doc_params or {})
```

tests/core/test_callable_utils.py
```diff
@@ -170,10 +176,12 @@ def func_with_extra_param_in_docstring(a: int) -> str:
         return str(a)
 
     # We expect the generated SQL to ignore 'b' since it's not in the function signature
-    sql_body = generate_sql_function_body(
-        func_with_extra_param_in_docstring, "test_catalog", "test_schema"
-    )
+    with pytest.warns(UserWarning) as record:
```
harupy:
```suggestion
    with pytest.warns(UserWarning, match="<message>") as record:
```

We should always use match when using `pytest.warns`. Imagine your insert `warnings.warn("test", UserWarning)` for testing and forgot to remove it.

tests/core/test_callable_utils.py
BenWilson2: I've always liked the standard library recorder for this stuff (makes the tests shorter)

tests/core/test_callable_utils.py
BenWilson2: I shifted over to pytest.warns since it's the standard we use in MLflow. The only reason why I don't like using it for these cases is that it's WAY more verbose for zero gain. 
For sake of consistency and familiarity, migrated to pytest.

tests/core/test_callable_utils.py
BenWilson2: ? Isn't that negated by the reference alias? 

The official docs for pytest show this exact example and also explain when match is to be used (capturing a specific single warning regex match statement), which for code readability, I externalized to parse the recorded warnings. 

https://docs.pytest.org/en/stable/how-to/capture-warnings.html#warns 

I can add the match with a wildcard, but that's a no-op regex match all (which is the same as the default behavior)

src/ucai/core/utils/callable_utils.py
```diff
@@ -428,3 +432,50 @@ def validate_return_type(func_name: str, type_hints: dict[str, Any]) -> str:
...
+    Args:
+        doc_params (Optional[dict[str, str]]): Parameters documented in the docstring.
+        signature_params (Set[str]): Parameters present in the function signature.
+        func_name (str): The name of the function being checked.
+
+    Returns:
+        None
+    """
+    params_in_doc = set(doc_params.keys() or {})
+    params_in_signature = signature_params
```
harupy: do we need  this reassignment?

BenWilson2: nope, removed

tests/core/test_callable_utils.py
harupy: I'm a bit lost. `with pytest.warns(UserWarning, match="The following parameters are documented in the docstring but not present in the function signature: b")` doesn't work?

BenWilson2: created an option match for the multiple capture scenarios. It's definitely a bit cleaner than parametrizing the warnings. 
If it becomes too annoying for changes in the future, we can look into parametrizing the matches.

PR #33, opened 2024-10-10
Address bug bash nits

serena-ruan:
Remove wrapper of exceptions;
Remove confusing function_info parameter in `create_function` as it's not supported for now. (we might reconsider about whether this needs to be supported)
Fix list_functions docstring return type.


harupy (approved):
LGTM

PR #34, opened 2024-10-10
Map python int to LONG in SQL

serena-ruan:
INTEGER has limitation of 4-byte signed integer numbers, but python int can be larger than that, so we map python int to SQL LONG type, which is 8-byte signed integer number (biggest range SQL can hold)

### DIFF
--- a/tests/core/test_callable_utils.py
+++ b/tests/core/test_callable_utils.py
@@ -432,7 +432,7 @@ def test_function_with_multiple_return_paths():
     sql_body = generate_sql_function_body(multiple_return_func, "test_catalog", "test_schema", True)
 
     expected_sql = """
-CREATE OR REPLACE FUNCTION test_catalog.test_schema.multiple_return_func(a LONG COMMENT 'Parameter a')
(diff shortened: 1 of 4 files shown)


BenWilson2 (approved):
Good catch!

tests/core/test_callable_utils.py
```diff
@@ -432,7 +432,7 @@ def multiple_return_func(a: int) -> str:
     sql_body = generate_sql_function_body(multiple_return_func, "test_catalog", "test_schema", True)
 
     expected_sql = """
-CREATE OR REPLACE FUNCTION test_catalog.test_schema.multiple_return_func(a INTEGER COMMENT 'An integer')
+CREATE OR REPLACE FUNCTION test_catalog.test_schema.multiple_return_func(a LONG COMMENT 'Parameter a')
```
BenWilson2:
```suggestion
CREATE OR REPLACE FUNCTION test_catalog.test_schema.multiple_return_func(a LONG COMMENT 'An integer')
```

PR #35, opened 2024-10-10
Address edge cases in docstring parsing

BenWilson2:
This is intended to fix several edge cases with arg and return naming conventions (added tests to cover these use cases)


src/ucai/core/utils/docstring_utils.py
```diff
@@ -0,0 +1,181 @@
...
+        raise ValueError(
+            "Docstring is empty. Please provide a docstring with a function description."
+        )
+
+    description_lines = []
+    parsed_params = {}
+    returns = None
+    current_param = None
+    param_description_lines = []
+    return_description_lines = []
```
harupy: I'd recommend adding type hints if we do "initialize-then-mutate".

harupy:
```suggestion
    description_lines: list[str] = []
    parsed_params: dict[str, str] = {}
    returns: str | None = None
    current_param str | None = None
    param_description_lines: list[str] = []
    return_description_lines: list[str] = []
```

tests/core/test_docstring_utils.py
```diff
@@ -0,0 +1,237 @@
...
+        int: The input value incremented by one.
+    """
+    with pytest.raises(ValueError, match="Function description is missing"):
+        parse_docstring(docstring)
+
+
+def test_parse_docstring_single_line_single_param():
+    """
+    Test parsing a simple docstring with a single-line description and one parameter.
+    """
```
harupy: Do we need this docstring? I think the function name is descriptive enough.

tests/core/test_docstring_utils.py
```diff
@@ -0,0 +1,237 @@
...
+
+from ucai.core.utils.docstring_utils import DocstringInfo, parse_docstring
+
+
+def test_parse_docstring_empty():
+    """
+    Test that an empty docstring raises a ValueError.
+    """
+    with pytest.raises(ValueError, match="Docstring is empty"):
+        parse_docstring("")
```
harupy:
```suggestion
def test_parse_docstring_empty_raises_value_error():
    with pytest.raises(ValueError, match="Docstring is empty"):
        parse_docstring("")
```

tests/core/test_docstring_utils.py
harupy: docstring might make sense for a complex test case, but I don't think we need docstring for simple test cases.

tests/core/test_docstring_utils.py
```diff
@@ -0,0 +1,237 @@
...
+def test_parse_docstring_google_style():
+    """
+    Test parsing a Google-style docstring.
+    """
+    docstring = """
+    Calculate the sum of two numbers.
+
+    Args:
+        a (int): The first number.
+        b (int): The second number.
```
harupy, line 100:
```suggestion
        a (int): The first number.

        b (int): The second number.
```

what would happen if there is a blank line between a and b? Is this case handled?

tests/core/test_docstring_utils.py
```diff
@@ -0,0 +1,237 @@
...
+    Process data.
+
+    Args:
+        data:
+        config:
+    Returns:
+        bool: Success status.
+    """
+    expected = DocstringInfo(
+        description="Process data.", params={}, returns="bool: Success status."
```
harupy: shouldn't params be `{"a": "", "b": ""}` or `{"a": None, "b": None}`?

tests/core/test_docstring_utils.py
```diff
@@ -0,0 +1,237 @@
...
+    assert result == expected
+
+
+def test_parse_docstring_no_params():
+    """
+    Test parsing a docstring with no parameters.
+    """
+    docstring = """
+    Get the current timestamp.
+
```
serena-ruan, line 81: Can we also add test for
```
Args:
Returns:...
```

tests/core/test_docstring_utils.py
BenWilson2: Removed them all. The titles are explanatory enough

tests/core/test_docstring_utils.py
BenWilson2: this worked, but added tests that cover this

tests/core/test_docstring_utils.py
BenWilson2: yep, adjusted accordingly

tests/core/test_docstring_utils.py
BenWilson2: yep, this correctly parses now (added tests too)

src/ucai/core/utils/docstring_utils.py
```diff
@@ -0,0 +1,154 @@
...
+                current_section = section
+                content = stripped_line[len(title) :].strip()
+                if content:
+                    sections[current_section].append(content)
+                matched_section = True
+                break
+        if matched_section:
+            continue
+        sections[current_section].append(line)
+    return sections
```
harupy, line 74: Not a blocker, but I'd return a dataclass.

harupy (approved):
LGTM

serena-ruan (approved):
Thanks for the fix! Let's port this over to unitycatalog repo :D

PR #36, opened 2024-10-11
Add repr for Llamaindex UnitycatalogTool

serena-ruan:


integrations/llama_index/src/ucai_llamaindex/toolkit.py
```diff
@@ -41,6 +42,13 @@ def __init__(
         super().__init__(*args, fn=fn, metadata=metadata, **kwargs)
         self.client_config = client_config
 
+    def __repr__(self) -> str:
+        return (
+            "UnityCatalogTool("
+            + ", ".join([f"{k}={v}" for k, v in asdict(self.metadata).items() if v])
```
harupy: why `if v`? Should we ignore an empty string

harupy:
```suggestion
            + ", ".join([f"{k}={v!r}" for k, v in asdict(self.metadata).items() if v])
```

to quote string values.

harupy:
```suggestion
            + ", ".join(f"{k}={v}" for k, v in asdict(self.metadata).items() if v)
```

to skip temporary list creation.

serena-ruan: We can include empty string :) Let me update

harupy (approved):
LGTM

PR #37, opened 2024-10-11
Fix extract_function_name

serena-ruan:
The fix is to make sure the function name strictly follows the syntax in the sql body and we correctly extract them.

### DIFF
--- a/src/ucai/core/databricks.py
+++ b/src/ucai/core/databricks.py
@@ -25,7 +25,11 @@ from ucai.core.utils.type_utils import (
     convert_timedelta_to_interval_str,
     is_time_type,
 )
-from ucai.core.utils.validation_utils import validate_full_function_name, validate_param
+from ucai.core.utils.validation_utils import (
+    strip_backticks,
+    validate_full_function_name,
+    validate_param,
+)
 
 if TYPE_CHECKING:
     from databricks.sdk import WorkspaceClient
@@ -104,7 +108,14 @@ def extract_function_name(sql_body: str) -> str:
 
     match = pattern.search(sql_body)
     if match:
-        return match.group(1)
+        result = match.group(1)
+        full_function_name = validate_full_function_name(result)
+        # backticks are only required in SQL, not in python APIs
+        return (
+            f"{strip_backticks(full_function_name.catalog_name)}."
+            f"{strip_backticks(full_function_name.schema_name)}."
+            f"{strip_backticks(full_function_name.function_name)}"
+        )
     raise ValueError(
         f"Could not extract function name from the sql body: {sql_body}.\nPlease "
         "make sure the sql body follows the syntax of CREATE FUNCTION "
--- a/src/ucai/core/databricks.py
+++ b/src/ucai/core/databricks.py
@@ -369,7 +369,7 @@ class DatabricksFunctionClient(BaseFunctionClient):
             FunctionInfo: The function info.
         """
         full_func_name = validate_full_function_name(function_name)
-        if "*" in full_func_name.function_name:
+        if "*" in full_func_name.function:
             raise ValueError(
                 "function name cannot include *, to get all functions in a catalog and schema, "
                 "please use list_functions API instead."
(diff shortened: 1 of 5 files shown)


serena-ruan:
https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/11287027455 passed

src/ucai/core/databricks.py
```diff
@@ -89,22 +93,31 @@ def extract_function_name(sql_body: str) -> str:
...
     if match:
-        return match.group(1)
+        result = match.group(1)
+        full_function_name = validate_full_function_name(result)
+        # backticks are only required in SQL, not in python APIs
+        return (
+            f"{strip_backticks(full_function_name.catalog_name)}."
+            f"{strip_backticks(full_function_name.schema_name)}."
+            f"{strip_backticks(full_function_name.function_name)}"
+        )
```
harupy: I'd add a method to `FullFunctionName`. Imagine you need to do the same thing in other places.

src/ucai/core/utils/validation_utils.py
```diff
@@ -29,6 +29,12 @@ def validate_full_function_name(function_name: str) -> FullFunctionName:
     return FullFunctionName(catalog_name=splits[0], schema_name=splits[1], function_name=splits[2])
 
 
+def strip_backticks(s: str) -> str:
+    if s.startswith("`") and s.endswith("`"):
+        return s[1:-1]
+    return s
```
harupy: do we need this function? can we just use ``strip("`")``?

src/ucai/core/databricks.py
harupy: can we strip `_name` in each field? That seems redundant.

harupy (approved):
LGTM

src/ucai/core/utils/function_processing_utils.py
```diff
@@ -101,7 +101,7 @@ def get_tool_name(func_name: str) -> str:
     # OpenAI has constriant on the function name:
     # Must be a-z, A-Z, 0-9, or contain underscores and dashes, with a maximum length of 64.
     full_func_name = validate_full_function_name(func_name)
-    tool_name = f"{full_func_name.catalog_name}__{full_func_name.schema_name}__{full_func_name.function_name}"
+    tool_name = f"{full_func_name.catalog}__{full_func_name.schema}__{full_func_name.function}"
```
harupy: can we make this as a method of `FullFunctionName`?

src/ucai/core/databricks.py
```diff
@@ -89,22 +92,27 @@ def extract_function_name(sql_body: str) -> str:
...
     """
+    # NOTE: catalog/schema/function names follow guidance here:
+    # https://docs.databricks.com/en/sql/language-manual/sql-ref-names.html#catalog-name
     pattern = re.compile(
         r"""
         CREATE\s+(?:OR\s+REPLACE\s+)?      # Match 'CREATE OR REPLACE' or just 'CREATE'
         (?:TEMPORARY\s+)?                  # Match optional 'TEMPORARY'
         FUNCTION\s+(?:IF\s+NOT\s+EXISTS\s+)?  # Match 'FUNCTION' and optional 'IF NOT EXISTS'
-        ([\w.]+)                           # Capture the function name (including schema if present)
+        ([^ /.]+\.[^ /.]+\.[^ /.]+)          # Capture the function name (including schema if present)
```
harupy: In a follow-up, can we use named groups: https://docs.python.org/3.12/library/re.html?


`group(1)`: where is the first group?
`group("name")`: can easily locate where the name group is.

src/ucai/core/utils/validation_utils.py
```diff
@@ -6,9 +6,12 @@
...
-    function_name: str
+    catalog: str
+    schema: str
+    function: str
+
+    def __str__(self) -> str:
+        return f"{self.catalog.strip('`')}.{self.schema.strip('`')}.{self.function.strip('`')}"
 
 
 def validate_full_function_name(function_name: str) -> FullFunctionName:
```
harupy: can we make this a class method of `FullFunctionName`?

PR #39, opened 2024-10-11
Add client session refresh logic

BenWilson2:
Adds retry logic for the two primary mechanism of token expiry and refreshes the connections that the client may make to re-fetch the auth tokens needed. 
Services mocked in tests.

### DIFF
--- a/src/ucai/core/databricks.py
+++ b/src/ucai/core/databricks.py
@@ -52,6 +52,7 @@ DATABRICKS_CONNECT_VERSION_NOT_SUPPORTED_ERROR_MESSAGE = (
 )
 SESSION_RETRY_BASE_DELAY = 1
 SESSION_RETRY_MAX_DELAY = 32
+SESSION_EXCEPTION_MESSAGE = "session_id is no longer usable"
 
 _logger = logging.getLogger(__name__)
 
@@ -88,6 +89,15 @@ def _try_get_spark_session_in_dbr() -> Any:
         return
 
 
+def _is_in_databricks_notebook_environment() -> bool:
+    try:
+        from dbruntime.databricks_repl_context import get_context
+
+        return get_context().isInNotebook()
+    except (ImportError, AttributeError):
+        return False
+
+
 def extract_function_name(sql_body: str) -> str:
     """
     Extract function name from the sql body.
@@ -115,53 +125,44 @@ def extract_function_name(sql_body: str) -> str:
     )
 
 
-def retry_on_session_expiration():
+def retry_on_session_expiration(func):
     """
     Decorator to retry a method upon session expiration errors with exponential backoff.
-
-    Args:
-        max_attempts (int): Maximum number of retry attempts.
-        base_delay (int): Base delay in seconds for exponential backoff.
-        max_delay (int): Maximum delay in seconds between retries.
     """
     max_attempts = int(UCAI_DATABRICKS_SESSION_RETRY_MAX_ATTEMPTS.get())
 
-    def decorator(func):
-        @functools.wraps(func)
-        def wrapper(self, *args, **kwargs):
-            for attempt in range(1, max_attempts + 1):
-                try:
-                    return func(self, *args, **kwargs)
-                except Exception as e:
-                    error_message = str(e)
-                    if (
-                        "session_id is no longer usable" in error_message
-                        or "BAD_REQUEST: session_id is no longer usable" in error_message
-                    ):
-                        if attempt < max_attempts:
-                            delay = min(
-                                SESSION_RETRY_BASE_DELAY * (2 ** (attempt - 1)),
-                                SESSION_RETRY_MAX_DELAY,
-                            )
-                            _logger.warning(
-                                f"Session expired. Attempt {attempt} of {max_attempts}. Refreshing session and retrying after {delay} seconds..."
-                            )
-                            self.refresh_client_and_session()
-                            time.sleep(delay)
-                            continue
-                        else:
-                            _logger.error(
-                                f"Failed to execute {func.__name__} after {max_attempts} attempts."
-                            )
-                            raise RuntimeError(
-                                f"Failed to execute {func.__name__} after {max_attempts} attempts due to session expiration."
-                            ) from e
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        for attempt in range(1, max_attempts + 1):
+            try:
+                return func(self, *args, **kwargs)
+            except Exception as e:
+                error_message = str(e)
+                if SESSION_EXCEPTION_MESSAGE in error_message:
+                    if not self._is_default_client:
+                        refresh_message = f"Failed to execute {func.__name__} due to session expiration. Unable to automatically refresh session when using a custom client."
+                        _logger.error(refresh_message)
+                        raise RuntimeError(refresh_message) from e
+
+                    if attempt < max_attempts:
+                        delay = min(
+                            SESSION_RETRY_BASE_DELAY * (2 ** (attempt - 1)),
+                            SESSION_RETRY_MAX_DELAY,
+                        )
+                        _logger.warning(
+                            f"Session expired. Attempt {attempt} of {max_attempts}. Refreshing session and retrying after {delay} seconds..."
+                        )
+                        self.refresh_client_and_session()
+                        time.sleep(delay)
+                        continue
                     else:
-                        raise
+                        refresh_failure_message = f"Failed to execute {func.__name__} after {max_attempts} attempts due to session expiration."
+                        _logger.error(refresh_failure_message)
+                        raise RuntimeError(refresh_failure_message) from e
+                else:
+                    raise
 
-        return wrapper
-
-    return decorator
+    return wrapper
 
 
 class DatabricksFunctionClient(BaseFunctionClient):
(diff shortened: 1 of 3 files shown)


src/ucai/core/databricks.py
```diff
@@ -168,6 +221,14 @@ def _validate_warehouse_type(self):
                 "https://docs.databricks.com/en/admin/sql/serverless.html#enable-serverless-sql-warehouses."
             )
 
+    def refresh_client_and_session(self):
+        _logger.info("Refreshing Databricks client and Spark session due to session expiration.")
+        self.client = get_default_databricks_workspace_client(profile=self.profile)
```
serena-ruan, line 230: If the client is manually passed by the user, this might not work. 
Can we add a flag in __init__ to check whether user passed a client, and if so we include that info in the _logger.info and suggest them to re-initialize the client or update default auth so we can refresh automatically.

src/ucai/core/databricks.py
```diff
@@ -168,6 +221,14 @@ def _validate_warehouse_type(self):
...
+        self.stop_spark_session()
+        self.set_default_spark_session()
+        self.spark = _try_get_spark_session_in_dbr()
```
serena-ruan: I had a simple test, this logic seems broken if in databricks notebook, since we can't directly reinitialize the spark session (which requires some connection string) in notebook.
Can we check if we're in dbr first, and if so we do nothing to the spark session (or we can investigate to see how to refresh the cluster); and if we're not in dbr, we can stop + set again.

src/ucai/core/databricks.py
```diff
@@ -111,6 +115,55 @@ def extract_function_name(sql_body: str) -> str:
...
+        @functools.wraps(func)
+        def wrapper(self, *args, **kwargs):
+            for attempt in range(1, max_attempts + 1):
+                try:
+                    return func(self, *args, **kwargs)
+                except Exception as e:
+                    error_message = str(e)
+                    if (
+                        "session_id is no longer usable" in error_message
+                        or "BAD_REQUEST: session_id is no longer usable" in error_message
```
serena-ruan:
```suggestion
                        "session_id is no longer usable" in error_message
```
Do we need `or`?

src/ucai/core/databricks.py
```diff
@@ -111,6 +115,55 @@ def extract_function_name(sql_body: str) -> str:
...
+    Decorator to retry a method upon session expiration errors with exponential backoff.
+
+    Args:
+        max_attempts (int): Maximum number of retry attempts.
+        base_delay (int): Base delay in seconds for exponential backoff.
+        max_delay (int): Maximum delay in seconds between retries.
+    """
+    max_attempts = int(UCAI_DATABRICKS_SESSION_RETRY_MAX_ATTEMPTS.get())
+
+    def decorator(func):
```
serena-ruan: Can we remove this and directly return wrapper? Update the function to retry_on_session_expiration(func).

src/ucai/core/databricks.py
```diff
@@ -197,6 +258,7 @@ def create_function(
         # return self.client.functions.create(function_info)
         raise ValueError("sql_function_body must be provided.")
 
+    @retry_on_session_expiration()
```
serena-ruan: Do we need `()`? Can we use @retry_on_session_expiration

src/ucai/core/databricks.py
```diff
@@ -590,6 +654,11 @@ def _execute_uc_functions_with_serverless(
                     format="CSV", value=csv_buffer.getvalue(), truncated=truncated
                 )
         except Exception as e:
+            if (
+                "session_id is no longer usable"
```
serena-ruan: Can we make this sentence a constant?

BenWilson2:
Verification:

``` python
# MonkeyPatch spark.sql used in the execute function
original_sql_method = client_std.spark.__class__.sql

call_count = {'count': 0}

def side_effect(self, *args, **kwargs):
    call_count['count'] += 1
    if call_count['count'] == 1:
        # Simulate the session expiration exception
        raise Exception("session_id is no longer usable")
    else:
        if 'sqlQuery' in kwargs:
            sql_query = kwargs.pop('sqlQuery')
            return original_sql_method(self, sql_query, *args, **kwargs)
        else:
            return original_sql_method(self, *args, **kwargs)

client_std.spark.sql = side_effect.__get__(client_std.spark, client_std.spark.__class__)

# First call raises, second should execute
result = client_std.execute_function(
    function_name="ben_wilson.uc_func.weather", 
    parameters={"location": "tx"}
)

# Clean up the patch
client_std.spark.sql = original_sql_method.__get__(client_std.spark, client_std.spark.__class__)

# Validate
print(f"Number of calls to spark.sql: {call_count['count']}")
print(f"Function execution result: {result}")
```
Screenshot 2024-10-15 at 3 11 59 PM

src/ucai/core/databricks.py
BenWilson2: Good call here. We shouldn't attempt to retry if a client is being passed. That's up to the user to handle that. Updated!

src/ucai/core/databricks.py
BenWilson2: Yep - confirmed. It's completely broken in serverless (it also makes a fun time of requiring a termination / restart command if this method is called!). Added a check for interactive notebook validation, removal of spark session termination and relying on a refresh of the client instance for the WorkspaceClient and simply refreshing the reference to the notebook REPL SparkSession obj.

src/ucai/core/databricks.py
BenWilson2: Not sure what the heck I was thinking here? lol... long day.

src/ucai/core/databricks.py
BenWilson2: Yeah, I originally had the factory decorator for args passing of client-based overrides to retry logic. Forgot to remove this when moving to the ENV VAR approach. Cleaned up!

src/ucai/core/databricks.py
BenWilson2: It was needed for the factory approach, but after removing that additional layer that supports arg passing, it's no longer needed. Removed!

src/ucai/core/databricks.py
BenWilson2: indeed :)

src/ucai/core/databricks.py
```diff
@@ -111,6 +125,46 @@ def extract_function_name(sql_body: str) -> str:
...
+    def wrapper(self, *args, **kwargs):
+        for attempt in range(1, max_attempts + 1):
+            try:
+                return func(self, *args, **kwargs)
+            except Exception as e:
+                error_message = str(e)
+                if SESSION_EXCEPTION_MESSAGE in error_message:
+                    if not self._is_default_client:
+                        refresh_message = f"Failed to execute {func.__name__} due to session expiration. Unable to automatically refresh session when using a custom client."
+                        _logger.error(refresh_message)
```
serena-ruan: Do we need this? refresh_message is included in the RuntimeError

src/ucai/core/databricks.py
```diff
@@ -111,6 +125,46 @@ def extract_function_name(sql_body: str) -> str:
...
+                        )
+                        _logger.warning(
+                            f"Session expired. Attempt {attempt} of {max_attempts}. Refreshing session and retrying after {delay} seconds..."
+                        )
+                        self.refresh_client_and_session()
+                        time.sleep(delay)
+                        continue
+                    else:
+                        refresh_failure_message = f"Failed to execute {func.__name__} after {max_attempts} attempts due to session expiration."
+                        _logger.error(refresh_failure_message)
```
serena-ruan: Same here, seems no need

src/ucai/core/databricks.py
```diff
@@ -84,6 +89,15 @@ def _try_get_spark_session_in_dbr() -> Any:
         return
 
 
+def _is_in_databricks_notebook_environment() -> bool:
```
serena-ruan, line 92: This is a bit different than https://github.com/mlflow/mlflow/blob/7b16d762d21b44791f097ffaae101a9dc7600a53/mlflow/utils/databricks_utils.py#L175 
But as long as this field is enough for checking `in notebook` then we are good!

src/ucai/core/databricks.py
```diff
@@ -84,6 +89,15 @@ def _try_get_spark_session_in_dbr() -> Any:
...
+    try:
+        from dbruntime.databricks_repl_context import get_context
+
+        return get_context().isInNotebook
+    except (ImportError, AttributeError):
```
serena-ruan: What about other exceptions 🤔 This is internal function so if there're other exceptions there's nothing users can do. Maybe we just always return False when there's exception?

serena-ruan (approved):
Mostly LGTM! Thanks for the validation!

src/ucai/core/databricks.py
BenWilson2: Yeah the logic within MLflow is a bit intense (it has to be to support usage in so many different runtime environments). For serverless notebooks and classic compute, this simple check should be sufficient for development work of users. 
We may end up revisiting this if users start trying to do jobs-based notebook execution of dynamic UC function creation, but until someone complains about that, I think we should leave this as lightweight as possible :)

src/ucai/core/databricks.py
BenWilson2: Excellent point! broadened the scope

src/ucai/core/databricks.py
BenWilson2: As a final statement at the end of an unrecoverable error, logging makes zero sense :) removed!

PR #41, opened 2024-10-14
Fix langchain toolkit fields validation

serena-ruan:
Fix the 'keyerror' problem when passing a non-string param to function_names.


B-Step62 (approved):
LGTM! Thanks for the fix!

PR #43, opened 2024-10-15
Add delete_function API

serena-ruan:
Add delete_function API for clients.

### DIFF
--- a/src/ucai/core/client.py
+++ b/src/ucai/core/client.py
@@ -175,6 +175,7 @@ class BaseFunctionClient(ABC):
         Args:
             function_name: The full name of the function to delete.
                 It should be in the format of "catalog.schema.function_name".
+            kwargs: additional key-value pairs to include when deleting the function.
         """
 
     @abstractmethod
--- a/src/ucai/core/client.py
+++ b/src/ucai/core/client.py
@@ -164,6 +164,20 @@ class BaseFunctionClient(ABC):
         """
 
     @abstractmethod
+    def delete_function(
+        self,
+        full_function_name: str,
+        **kwargs,
+    ) -> None:
+        """
+        Delete a function by its full name.
+
+        Args:
+            full_function_name: The full name of the function to delete.
+                It should be in the format of "catalog.schema.function_name".
+        """
+
+    @abstractmethod
     def to_dict(self):
         """
         Store the client configuration in a dictionary.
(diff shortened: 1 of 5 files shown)


src/ucai/core/databricks.py
```diff
@@ -602,6 +602,22 @@ def _execute_uc_functions_with_serverless(
...
+        full_function_name: str,
+        force: Optional[bool] = None,
+    ) -> None:
+        """
+        Delete a function by its full name.
+
+        Args:
+            full_function_name: The full name of the function to delete.
+                It should be in the format of "catalog.schema.function_name".
+            force: Force deletion even if the function is notempty. Defaults to None.
```
harupy: what does None mean?

harupy:
```suggestion
            force: Force deletion even if the function is not empty. Defaults to None.
```

src/ucai/core/client.py
```diff
@@ -163,6 +163,20 @@ def _execute_uc_function(
         Internal logic for executing a UC function.
         """
 
+    @abstractmethod
+    def delete_function(
+        self,
+        full_function_name: str,
```
harupy: do we use `full_function_name` in other methods? We should be consistent.

src/ucai/core/databricks.py
serena-ruan: Honestly I don't know :) The underlying sdk uses default value None, if it's none then it doesn't pass this field to the request body

src/ucai/core/client.py
```diff
@@ -163,6 +163,20 @@ def _execute_uc_function(
...
+        self,
+        function_name: str,
+        **kwargs,
+    ) -> None:
+        """
+        Delete a function by its full name.
+
+        Args:
+            function_name: The full name of the function to delete.
+                It should be in the format of "catalog.schema.function_name".
```
harupy, line 177:
```suggestion
                It should be in the format of "catalog.schema.function_name".
            kwargs: ...
```

to explain what kwargs is.

src/ucai/core/databricks.py
harupy: > The underlying sdk uses default value None

I think we can mention this if the underlying client decides what to do with None.

harupy (approved):
LGTM

